PR #17, opened 2021-01-21
Add pytest ini options to config

jbusecke:


jbusecke:
We should merge this too, since #17 s instructions rely on those options.

Issue #18, opened 2021-02-02
improve testing of Filter class


rabernat:
I implemented the filter class based on @iangrooms's original code. The filter class is responsible for picking the coefficients for the biharmonic and laplacian steps. It is tested in this file:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/master/tests/test_filter.py

However, it would be best if someone who understands the procedure for defining the filter more deeply would help improve these tests. Right now I am just checking that a particular set of input parameters produces the same outputs I got in my own development environment, e.g.

https://github.com/ocean-eddy-cpt/gcm-filters/blob/04ad83562a670db8d194d7ad672059d74197a109/tests/test_filter.py#L20-L30

Is this sensible? Or can we think of more fundamental ways to test this part of the code.

rabernat:
- Check that filter weights give good approx of target polynomial
- Check zero crossing

rabernat:
As discussed in #20, we have a small gap in test coverage for different values of `ndim` here: https://codecov.io/gh/ocean-eddy-cpt/gcm-filters/pull/20/src/gcm_filters/filter.py?before=gcm_filters/filter.py#L70...82

PR #20, opened 2021-02-05
Adding default n_steps and adding ndim

iangrooms:
This PR does a few things:
1. Add in a default for `n_steps` that depends on the coarsening factor (`filter_scale/dx_min`)
2. Add an option to switch between 1d and 2d grids `ndmin`
3. Change the meaning of `filter_scale` in both filter shapes, so that it always corresponds to a boxcar of width `filter_scale`

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -1,12 +1,12 @@
 """Main Filter class."""
 import enum
 
-from dataclasses import dataclass, field
 from typing import Iterable, NamedTuple
 
 import numpy as np
 import xarray as xr
 
+from dataclasses import dataclass, field
 from scipy import interpolate
 
 from .gpu_compat import get_array_module
@@ -30,25 +30,18 @@ def _gaussian_target(target_spec: TargetSpec):
 
 
 def _taper_target(target_spec: TargetSpec):
-    return interpolate.PchipInterpolator(
+    FK = interpolate.PchipInterpolator(
         np.array(
             [
-                -1,
-                (2 / target_spec.s_max)
-                * (
-                    2
-                    * np.pi
-                    / (target_spec.transition_width * target_spec.filter_scale)
-                )
-                ** 2
-                - 1,
-                (2 / target_spec.s_max) * (2 * np.pi / target_spec.filter_scale) ** 2
-                - 1,
-                2,
+                0,
+                2 * np.pi / (target_spec.transition_width * target_spec.filter_scale),
+                2 * np.pi / target_spec.filter_scale,
+                2 * np.sqrt(target_spec.s_max),
             ]
         ),
         np.array([1, 1, 0, 0]),
     )
+    return lambda t: FK(np.sqrt((t + 1) * (target_spec.s_max / 2)))
 
 
 _target_function = {
@@ -218,8 +211,8 @@ class Filter:
 
     def __post_init__(self):
 
-        if self.n_steps <= 2:
-            raise ValueError("Filter requires N>2")
+        if self.n_steps < 0:
+            raise ValueError("Filter requires N>=0")
 
         self.filter_spec = _compute_filter_spec(
             self.filter_scale,
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -189,7 +189,8 @@ class Filter:
           Scales larger than ``pi * filter_scale / 2`` are left as-is. In between is a smooth transition.
     transition_width : float, optional
         Width of the transition region in the "Taper" filter.
-    ndim : Laplacian is applied on a grid of dimension ndim
+    ndim : int, optional
+         Laplacian is applied on a grid of dimension ndim
     grid_type : GridType
         what sort of grid we are dealing with
     grid_vars : dict
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -1,12 +1,12 @@
 """Main Filter class."""
 import enum
 
+from dataclasses import dataclass, field
 from typing import Iterable, NamedTuple
 
 import numpy as np
 import xarray as xr
 
-from dataclasses import dataclass, field
 from scipy import interpolate
 
 from .gpu_compat import get_array_module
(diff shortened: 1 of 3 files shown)


rabernat:
Thanks for getting this started @iangrooms! Let me know when you're ready for some feedback.

The "pre-commit" part of the checks does code linting (checking that the code formatting conforms to standards, etc.) You can set it up locally following the [contributor guide](https://gcm-filters.readthedocs.io/en/latest/how_to_contribute.html#linting-and-docstrings) and have it run automatically every time you commit. It will give hints on how to satisfy its checks.

iangrooms:
@rabernat this pull request should be ready. Not sure why it says it's failing the pre-commit, since it passes on my machine. Also, all tests currently pass.

gcm_filters/filter.py
```diff
@@ -60,8 +58,27 @@ class FilterSpec(NamedTuple):
...
+        if filter_shape == FilterShape.GAUSSIAN:
+            if ndim == 1:
+                n_steps = np.ceil(1.3 * filter_scale / dx_min).astype(int)
+            else:  # ndim==2
+                n_steps = np.ceil(1.8 * filter_scale / dx_min).astype(int)
+        else:  # Taper
+            if ndim == 1:
+                n_steps = np.ceil(4.5 * filter_scale / dx_min).astype(int)
+            else:  # ndim==2
+                n_steps = np.ceil(6.4 * filter_scale / dx_min).astype(int)
```
rabernat, line 80: What if the user wants to do 3D `(ndim == 3)`? Is it not supported? Currently if someone passes `ndim=3` they will get a pretty undecipherable error.

gcm_filters/filter.py
```diff
@@ -165,12 +182,14 @@ class Filter:
...
     n_steps : int, optional
         Number of total steps in the filter
+        n_steps == 0 means the number of steps is chosen automatically
     filter_shape : FilterShape
         - ``FilterShape.GAUSSIAN``: The target filter has kernel :math:`e^{-|x/Lf|^2}`
         - ``FilterShape.TAPER``: The target filter has target grid scale Lf. Smaller scales are zeroed out.
           Scales larger than ``pi * filter_scale / 2`` are left as-is. In between is a smooth transition.
     transition_width : float, optional
         Width of the transition region in the "Taper" filter.
+    ndim : Laplacian is applied on a grid of dimension ndim
```
rabernat, line 192:
```suggestion
    ndim : int, optional
         Laplacian is applied on a grid of dimension ndim
```

gcm_filters/filter.py
```diff
@@ -165,12 +182,14 @@ class Filter:
         The smallest grid spacing. Should have same units as ``filter_scale``
     n_steps : int, optional
         Number of total steps in the filter
+        n_steps == 0 means the number of steps is chosen automatically
```
rabernat, line 185:
```suggestion
        ``n_steps == 0`` means the number of steps is chosen automatically
```

rabernat (commented):
This is great @iangrooms! Congrats on successfully jumping through all of the hoops (tests, pre-commit, etc.)

I don't understand the pre-commit failure either, but for now I think we can safely ignore it. 

I have a few tiny suggestions for docstring formatting and a question about how to handle the `ndim` parameter.

gcm_filters/filter.py
iangrooms: The only part not supported would be setting a default `n_steps`. I could add an error that you have to supply your own `n_steps` if you want to use `ndim` other than 1 or 2.

rabernat:
Ok great! 

Do you see anything in the [tutorial](https://gcm-filters.readthedocs.io/en/latest/tutorial.html) that needs to be updated because of the API change? If not, I think this is ready to merge.

iangrooms:
> Do you see anything in the tutorial that needs to be updated because of the API change? If not, I think this is ready to merge.

I checked the tutorial, and there shouldn't be any changes. I also added an error if the user tries to use `ndim>2` with the default `n_steps`. So I think it should be ready

rabernat:
Ok great, I'll merge. Thanks a lot @iangrooms!

For reference, the coverage report is referring to the fact that some parts of the new if / else block for `ndim` are not covered by tests (red lines here): https://codecov.io/gh/ocean-eddy-cpt/gcm-filters/pull/20/src/gcm_filters/filter.py?before=gcm_filters/filter.py#L70...82

To keep test coverage high, technically we should add test scenarios for each of these different options, including the errors. However, that feels a little too strict for the early stage this package is at, so we will let it slide for now. I'll put a reminder in #18 about this.

rabernat:
@jbusecke it won't let me merge because of the pre-commit failure. Can you look at this and try to figure out what's going on? Ian says the pre-commit passes locally, but the workflow is still failing. If the pre-commit is going to be flaky like this, I feel like we should disable it as a required check.

jbusecke:
Let me look into this.

jbusecke:
I can reproduce the error locally on my machine. @iangrooms could you confirm which conda environment you used? Did you run `pre-commit run --all-files` from the root directory?

Another thing I noticed is that you started this PR from your master branch. Not sure if that is causing this trouble though. In general it is recommended to check out a new branch with `git checkout -b <new_cool_branchname>`.

jbusecke:
Oh I see this was solved. What was the culprit in the end?

iangrooms:
pre-commit changes files, including ones that I wasn't working on (`kernels.py`). So when I ran pre-commit it would fail, then I'd make some changes, then it would pass. But I didn't commit the changes that pre-commit made to files that I wasn't working on. That's why pre-commit passed locally but failed on the server; once I committed the changes to kernels.py (not made by me), it fixed the problem on the server.

rabernat:
> once I committed the changes to kernels.py (not made by me), it fixed the problem on the server.

Yet strangely, kernels.py does not show up in the [files changed](https://github.com/ocean-eddy-cpt/gcm-filters/pull/20/files) tab of this PR. 🧐

I'm still learning about pre-commit myself, so it's useful to get this feedback.

jbusecke:
Thanks for the feedback @iangrooms. That is quite strange, actually. It should not let you push the commit if there are problems with the committed files. Or at least that is how I understand it. 

For what its worth, I personally always do a manual pre-commit run `pre-commit run --all-files`, then another `git add -u` before the commit I want to push, but that is maybe not the smoothest way of doing it.

Issue #21, opened 2021-02-10
Issue with pre-commit or mypy


NoraLoose:
I am ready to submit a pull request, but I am having issues with pre-commit, specifically mypy.

```
(gcm-filters-env) noraloose:~/gcm-filters> pre-commit run --all-files

[WARNING] The 'rev' field of repo 'https://github.com/psf/black' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.
Trim Trailing Whitespace.................................................Passed
Check python ast.........................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Check docstring is first.................................................Passed
Check for added large files..............................................Passed
Fix requirements.txt.....................................................Passed
File Contents Sorter.....................................................Passed
flake8...................................................................Passed
isort....................................................................Passed
seed isort known_third_party.............................................Passed
black....................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

gcm_filters/filter.py:55: error: syntax error in type comment
gcm_filters/filter.py:57: error: syntax error in type comment
Found 2 errors in 1 file (checked 8 source files)
```

This points me to 2 lines in filter.py, even though I did not touch that file.
https://github.com/ocean-eddy-cpt/gcm-filters/blob/73baa9ee7d8f0a3f482fa2a04e1c12b39bda0a5d/gcm_filters/filter.py#L53-L57

I continued and ignored the filter.py file, since I only want to commit changes in kernel.py:
```
git add gcm_filters/kernels.py
git commit -m "move computation of wet_fac to post_init"

[WARNING] The 'rev' field of repo 'https://github.com/psf/black' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.
Trim Trailing Whitespace.................................................Passed
Check python ast.........................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Check docstring is first.................................................Passed
Check for added large files..............................................Passed
Fix requirements.txt.................................(no files to check)Skipped
File Contents Sorter.................................(no files to check)Skipped
flake8...................................................................Passed
isort....................................................................Passed
seed isort known_third_party.............................................Passed
black....................................................................Passed
[...]

PR #22, opened 2021-02-10
move computation of wet_fac to post_init

NoraLoose:
This PR does the following things:

In `CartesianLaplacianWithLandMask`
- move the computation of `fac` (now: `wet_fac`) to pre_init. This factor is independent of the field which the Laplacian is applied to. If we apply the Laplacian many times (as we will for the filtering operation), computing the factor only once will speed things up.
- delete 1 unneccesary line: `out = field.copy()`

Following @rabernat's recommendation in https://github.com/ocean-eddy-cpt/gcm-filters/issues/21#issuecomment-776910274, I also disabled the mypy check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,9 +39,9 @@ repos:
   - id: black
     language_version: python3
 
-- repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.770
-  hooks:
-  - id: mypy
-    exclude: docs/source/conf.py
-    args: [--ignore-missing-imports]
+#- repo: https://github.com/pre-commit/mirrors-mypy
+#  rev: v0.770
+#  hooks:
+#  - id: mypy
+#    exclude: docs/source/conf.py
+#    args: [--ignore-missing-imports]
--- a/gcm_filters/kernels.py
+++ b/gcm_filters/kernels.py
@@ -60,22 +60,24 @@ class CartesianLaplacianWithLandMask(BaseLaplacian):
 
     wet_mask: ArrayType
 
-    def __call__(self, field: ArrayType):
-        np = get_array_module(field)
+    def __post_init__(self):
+        np = get_array_module(self.wet_mask)
 
-        out = field.copy()  # is this necessary?
-        out = np.nan_to_num(field)  # is this necessary?
-        out = self.wet_mask * out
-
-        fac = (
+        self.wet_fac = (
             np.roll(self.wet_mask, -1, axis=-1)
             + np.roll(self.wet_mask, 1, axis=-1)
             + np.roll(self.wet_mask, -1, axis=-2)
             + np.roll(self.wet_mask, 1, axis=-2)
         )
 
+    def __call__(self, field: ArrayType):
+        np = get_array_module(field)
+
+        out = np.nan_to_num(field)  # set all nans to zero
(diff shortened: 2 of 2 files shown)


rabernat:
This is great, thanks for the update! Two questions.

- Do you know if this is actually faster? Have you tested that? Just curious...
- Have you verified that it works on GPU? That's something our test suite currently can't do, since we don't have a GPU inside the github workflow

NoraLoose:
Thanks for looking at this!
> Have you verified that it works on GPU? That's something our test suite currently can't do, since we don't have a GPU inside the github workflow

I just tested this. Yes, it works on GPU.

> Do you know if this is actually faster? Have you tested that? Just curious...

The answer is: I just assumed it would be faster. Turns out it takes pretty much the same time, see [this notebook](https://github.com/NoraLoose/gcm-filters/blob/preinit-wetfac-test/docs/gcm_filters_wetfac_test.ipynb). 

The good news is that gcm_filters is much faster on GPU than on CPU (as we saw before).
For the example in the notebook: 
**CPU 3min 27.0s
GPU 14.5s**

Since this new implementation is not faster, we can just disregard this PR. Sorry for the noise!

NoraLoose:
Having a closer look at the code in filter.py, I now understand why this PR doesn't lead to a large speedup.

In my example, I am filtering 1600 2D fields (100 time slices, 16 layers). For each of the 1600 fields, the Laplacian is applied 8 times (since I chose `n_steps = 8`). I initially thought that with my PR, wet_fac would be computed only once, instead of 1600 * 8 times. But this is not the case. With the following code structure, __post_init__() for the Laplacian (and the computation of wet_fac) is called 1600 times.

https://github.com/ocean-eddy-cpt/gcm-filters/blob/73baa9ee7d8f0a3f482fa2a04e1c12b39bda0a5d/gcm_filters/filter.py#L245-L253

In summary, while I expected to reduce 1600 * 8 computations to 1 computation, my PR led only to a reduction from 1600 * 8 to 1600 * 1 computations. Since you typically do not choose very high values for `n_steps` (you will actually run into numerical instabilities for large `n_steps`), I suggest that we just stick to the current implementation and close this PR.

rabernat:
> With the following code structure, **post_init**() for the Laplacian (and the computation of wet_fac) is called 1600 times.

Thanks for digging into this. It is extremely useful information. Did you run profiling to get it? If so, can you post the profile report here?

I think we should keep this PR and think of a way to refactor the kernel so that the initialization only has to be done once.

NoraLoose:
> Thanks for digging into this. It is extremely useful information. Did you run profiling to get it? If so, can you post the profile report here?

I haven't done any rigorous profiling, but saw no significant difference in the execution time. That, together with looking at the code structure, led me to this conclusion. But I could be wrong. 

I think I will prioritize the implementation of the Laplacian kernels for some of the gcm grids in the near future, but could get back to this performance question at a later time.

rabernat:
> That, together with looking at the code structure, led me to this conclusion

I think you are write. The kernel is initialized _inside_ the function passed to apply_ufunc:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/8b92035d4c57c7e387bbac3ed3f23e2a2c8c4bf2/gcm_filters/filter.py#L149-L156

We should be able to move the initialization (`Laplacian()`) _outside_.

Issue #9, opened 2021-01-12
Add POP Kernel


rabernat:
I actually already have one written and running on Casper

rabernat:
The package is ready to have laplacians added to it. They should be added to this file:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/master/gcm_filters/kernels.py

There are three things that need to be added in that file:
1. A new class (e.g. `POPTripolarLaplacian`), which inherits from `BaseLaplacian`. Any auxiliary grid variables (e.g. tarea, dx, dy, etc.) should be attributes of the class. The actual laplacian should be coded in the `__call__` method and have only a single argument.
2. Define a new `GridType` here:
   https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/gcm_filters/kernels.py#L14
   Maybe something like `POP_TRIPOLAR_GRID`
3. Then add the laplacian to the `ALL_KERNELS` dictionary, like this
   https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/gcm_filters/kernels.py#L14

Then the kernel will automatically get tested in the `test_kernels.py` module.
However, it probably won't work unless you define special options for the grid_vars, as here
https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/tests/test_kernels.py#L15-L18

For the tests, we will need to decide about #5. I have some ideas about that I'll share in that issue.

Issue #8, opened 2021-01-12
Add MOM5 kernel


rabernat:
cc @LaureZanna, as discussed on Slack

rabernat:
The package is ready to have laplacians added to it. They should be added to this file:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/master/gcm_filters/kernels.py

There are three things that need to be added in that file:
1. A new class (e.g. `MOM5Laplacian`), which inherits from `BaseLaplacian`. Any auxiliary grid variables (e.g. tarea, dx, dy, etc.) should be attributes of the class. The actual laplacian should be coded in the `__call__` method and have only a single argument.
2. Define a new `GridType` here:
   https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/gcm_filters/kernels.py#L14
   Maybe something like `MOM_GRID`
3. Then add the laplacian to the `ALL_KERNELS` dictionary, like this
   https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/gcm_filters/kernels.py#L14

Then the kernel will automatically get tested in the `test_kernels.py` module.
However, it probably won't work unless you define special options for the grid_vars, as here
https://github.com/ocean-eddy-cpt/gcm-filters/blob/91cd5794e1518e5631600b3590e173973c140291/tests/test_kernels.py#L15-L18

For the tests, we will need to decide about #5. I have some ideas about that I'll share in that issue.

cc @LaureZanna

rabernat:
@jbusecke maybe you could give a quick overview of the development process

- running tests locally
- pre-commit
- docstrings

ElizabethYankovsky:
@rabernat - Thanks for cc-ing! Laure, her postdoc Arthur, and I have been discussing the formulation of the Laplacian in MOM5 and will contribute to this

jbusecke:
Sure thing. Where should that go? As a docs page?

rabernat:
> Sure thing. Where should that go? As a docs page?

I was thinking just a comment on github, but a contributor guide on the docs is probably more robust. Could probably just be copy-pasted from one of our other projects.

jbusecke:
How detailed should this be with regard to git? Full step by step with add/commit/...?

jbusecke:
Take a look at my PR #16

LaureZanna:
[...]

Issue #1, opened 2021-01-11
Add MOM6 kernel


rabernat:
Right now we just have a trivial laplacian kernel:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/7b8d2906d0bdacb98846f4e39c3883ed3a3fc44a/gcm_filters/kernels.py#L17-L22

We need to add the kernel for MOM6. This should be based on whatever @NoraLoose is actively using in her notebooks.

NoraLoose:
@gustavo-marques, @sdbachman, and all other MOM6 pros - I am happy to hand this task over to you if you'd like to contribute to gcm-filters (which I heard you do :smiley: ).

The only reason why I am assigned to this issue is that I have filtering code for the NW2 (MOM6) domain. But I haven't done the global domain yet. Would be great to have your expertise here. I think I have some ideas for how to do the global Laplacian for MOM6, so I am happy to discuss these ideas (or leave it completely up to you if you prefer).  Just wanted to mention this, because I didn't want to give the false impression that I want to do all kernel implementations for all models. :nerd_face:

gustavo-marques:
Sounds good! Thanks, @NoraLoose. I will talk to @sdbachman and we will get back to you.

NoraLoose:
This issue was accidentally closed.

iangrooms:
We have both scalar and vector Laplacians for MOM6 now, so I'm closing this issue.

PR #23, opened 2021-02-11
Add Laplacian on irregular Cartesian grid

NoraLoose:
This PR adds a Laplacian on an irregular Cartesian grid: `IrregularCartesianLaplacianWithLandMask`. 

@iangrooms, @jakesteinberg: FYI, this is exactly the finite volume `Laplacian2D` (more specifically: the corrected version `Laplacian2D_FV`) that we started with, but
* works for a general grid with varying `dx`, `dy`, (i.e., does not assume constant `dy` and zonally constant `dx` as for our NW2 implementation)
* written much nicer, and works now with dask and on GPU.

The test that I added for this Laplacian is still on a **regular** grid (with dx=dy=1). I will follow up with a better test, where we test on an irregular grid (maybe a very simple lat-lon grid). This will also involve changing the `test_conservation` in `test_kernels.py` to an area-weighted integral conservation test.


NoraLoose:
My idea is that a lot of the gcm kernels can inherit from `IrregularCartesianLaplacianWithLandMask`.

As an example, the [POP kernel](https://github.com/ocean-eddy-cpt/gcm-filters/issues/9) (on a B grid) can use the `IrregularCartesianLaplacianWithLandMask` with the following choices:
* U fields (on corner points): `dxw` = `HTN`, `dyw` = `HUW`, `dxs` = `HUS`, `dys` = `HTE`,  `area` = `UAREA`.
* Tracer fields: `dxw` = np.roll(HUS, -1, axis=1), `dys` = np.roll(HTE, -1, axis=1), `dxs` = np.roll(HTN, -1, axis=0), `dys` = np.roll(HUW, -1, axis=0),  `area` = `TAREA`

Similarly, I fed the `IrregularCartesianLaplacianWithLandMask` with MOM6 grid variables (`dxw` `dyw` `dxs` `dys` `area`) in my tests, so I am linking [this issue](https://github.com/ocean-eddy-cpt/gcm-filters/issues/1), too. Something similar could be done for [MOM5,](https://github.com/ocean-eddy-cpt/gcm-filters/issues/8) too.

What would be left to do in all these cases, is the implementation of a proper handling of the (e.g., tripolar) boundary conditions.

rabernat:
In order for the new tutorial to show up in the docs, you need to link to it from the `index.rst` page.

NoraLoose:
@arthurBarthe, @ElizabethYankovsky - I updated the [tutorial notebook ](https://github.com/NoraLoose/gcm-filters/blob/cartesian-laplacian-irregular-grid/docs/tutorial_irregular_grid.ipynb) to illustrate how to filter with fixed filter length scale vs. fixed filter factor. I hope this helps! Let me know if anything is unclear.

rabernat:
I have some comments about your tutorial:

```python
st = xr.open_dataset('/glade/p/univ/unyu0004/neerajab/NeverWorld2/run%i/static.nc' % (run), decode_times=False) 
chunks = {'time': 20} 
av = xr.open_dataset('/glade/p/univ/unyu0004/neerajab/NeverWorld2/run%i/averages_000%i.nc' % (run,end_time-500+2), 
                     chunks=chunks, decode_times=False)
```

For the tutorials, I would be really nice to use public data, such as we can pull from an OPeNDAP server or from cloud storage. Of course I understand why your are doing it on Casper. But perhaps we could also point to the OSN data that Neeraja produced?

```python
dxw = xr.DataArray(data=st.dxCu.isel(xq=slice(0,Nx)),coords={'yh':av.yh,'xh':av.xh}, dims=('yh','xh'))
dyw = xr.DataArray(data=st.dyCu.isel(xq=slice(0,Nx)),coords={'yh':av.yh,'xh':av.xh}, dims=('yh','xh'))
dxs = xr.DataArray(data=st.dxCv.isel(yq=slice(0,Ny)),coords={'yh':av.yh,'xh':av.xh}, dims=('yh','xh'))
dys = xr.DataArray(data=st.dyCv.isel(yq=slice(0,Ny)),coords={'yh':av.yh,'xh':av.xh}, dims=('yh','xh'))
```

It's somewhat poor xarray style here to create a new DataArray out of an old one. I understand why you are doing this--because you need all the "grid_vars" to have the same size and dimensions. But I would prefer to use `rename` on the dimensions to accomplish this.

This actually highlights a big difference between what we are doing in this package and the [xgcm](https://xgcm.readthedocs.io/en/latest/grids.html) approach. With xgcm, we explicitly track the positions of the variables relative to the grid cell. Here, we are sweeping that under the rug, labeling everything to have the same dimensions, and letting the different kernels accept staggered variables. This could lead to confusion and errors.

An alternative would be to add xgcm as a dependency and use its machinery to handle staggered grids.

NoraLoose:
Thanks for the comments! 

> For the tutorials, I would be really nice to use public data, such as we can pull from an OPeNDAP server or from cloud storage.

I agree. I guess I mostly made this notebook to get other CPT members started (e.g., outlining how to do fixed factor vs. fixed scale filtering). But if we want to keep this tutorial beyond the development phase, I will look into the OSN data.

> It's somewhat poor xarray style here to create a new DataArray out of an old one. I understand why you are doing this--because you need all the "grid_vars" to have the same size and dimensions. But I would prefer to use rename on the dimensions to accomplish this.

Before choosing this "nuclear option", I actually started out with using `rename` for the dimensions. But I ran into issues because in addition to renaming, one also has to pass the values for the new coordinates. Xarray has probably an option to do this for you, which I will look into. Thanks!

> Here, we are sweeping that under the rug, labeling everything to have the same dimensions, and letting the different kernels accept staggered variables. This could lead to confusion and errors. An alternative would be to add xgcm as a dependency and use its machinery to handle staggered grids.

I agree. Not only do we let the kernels accept staggered variables, sometimes you have to find work-arounds to "trick" xarray and make it believe you are *not* on a staggered grid (see my previous comment). We should make it as easy as possible for the user. 

Is your idea to use xgcm only at the user level, e.g., in a fashion that I outlined in https://github.com/ocean-eddy-cpt/gcm-filters/issues/24#issuecomment-778903595? Under the hood, we would still have to bring all variables to the same dimensions to take full advantage of the generalized ufuncs + dask, correct?

rabernat:
> Is your idea to use xgcm only at the user level, e.g., in a fashion that I outlined in [#24 (comment)](https://github.com/ocean-eddy-cpt/gcm-filters/issues/24#issuecomment-778903595)? Under the hood, we would still have to bring all variables to the same dimensions to take full advantage of the generalized ufuncs + dask, correct?

Yes, I think we are thinking the same thing. What we would use from xgcm is its ability to "understand" grids. Xgcm knows if a variable is cell centered, left shift, right shift, etc. It also understands metrics like cell area, distance, etc.

We could use that high-level part to check and align the inputs, while keeping the low level of gcm-filters the same.

> Before choosing this "nuclear option", I actually started out with using `rename` for the dimensions. But I ran into issues because in addition to renaming, one also has to pass the values for the new coordinates.

I think we faced this in xgcm once too: see https://github.com/xgcm/xgcm/pull/153 for discussion.

rabernat:
I'd like to suggest a path forward so we can keep making progress.

### Short Term

Let's require, **for now**, that `field` and all `grid_vars` share the same dimensions (e.g. `nlat, nlon`). This will require the user to manually do what you are doing in some cases: rename / relabel the dimensions and coordinates of the grid vars. We recognize that this opens the door to potential user errors, such as misunderstanding what sort of inputs the laplacians require (e.g. cell centered, vorticity point, etc.) There is useful discussion on this in https://github.com/ocean-eddy-cpt/gcm-filters/issues/24. For now, we solve this problem with careful documentation

### Long Term

We refactor gcm-filters to leverage xgcm to "understand" the grid more deeply. Each laplacian would have to define the [grid positions](https://xgcm.readthedocs.io/en/latest/grids.html#axes-and-positions) of all its input variables. The grid_vars can be inferred from the xgcm [metrics](https://xgcm.readthedocs.io/en/latest/grid_metrics.html). @jbusecke does this sound doable?

If folks agree with this plan (leave a 👍 ), I think we can move forward and merge this PR.

NoraLoose:
You left some comments regarding the tutorial notebook that I submitted as part of this PR. If the plan is to add the notebook to the documentation (as suggested in this PR), I would like to put a bit more effort into the notebook, including fixing the things that you mentioned in your comments. I'm also okay with removing this notebook from the `index.rst` page until it is ready.

rabernat:
There were two main issues I identified in the tutorial
- Can we use public data? Yes, but we don't need to resolve this now
- Can we avoid that somewhat awkward way of relabeling the input dataarrays? This is basically unavoidable if we go with my "short term" suggestion above. So not sure there is much to do at this point.

> I'm also okay with removing this notebook from the `index.rst` page until it is ready.

At this point, we are our only "users" of this package, so we don't have to worry too much about what shows up on the documentation site.

So to summarize, if you want to continue tweak the tutorial, go ahead! But IMO it's good enough already to merge.

NoraLoose:
Cool. Let's merge it, and I will get back to the fine-tuning of the documentation once we have resolved some more fundamental issues in gcm-filters.

PR #25, opened 2021-02-11
Raise error message if filter_scale < dx_min

NoraLoose:


rabernat:
For a change like this, it would be great to add a test that covers the error, similar to this:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/8b92035d4c57c7e387bbac3ed3f23e2a2c8c4bf2/tests/test_filter.py#L102-L108

That will help keep the code coverage high.

It could live in a new standalone test, e.g. `test_filter_errors()`.

NoraLoose:
Thanks @rabernat, this makes sense. 

Before adding such a test, I would like to discuss an alternative option to handle `dx_min` (which would require different testing). 

The idea is to remove `dx_min` from the filter arguments, and instead
* compute `dx_min` in a separate function inside `filter.py`, if the user wants to filter with **fixed filter length scale**. `dx_min` would then be computed as the minimum grid spacing, inferred from the grid variables that the user puts in;
* set `dx_min` to 1, if the user wants to filter with **fixed coarsening factor**.

This could avoid errors and confusion on the user side. For instance, I anticipate that users will try to set `dx_min = 1`, even though they want to filter with **fixed filter length scale**. (I think @arthurBarthe ran into this issue at first.) If we decide to go this way, we should probably add a filter argument, via which the user picks a "method": **fixed filter length scale** or **fixed coarsening factor**. What do you think about this? I'm tagging @iangrooms here, since he has written most of the FilterSpec class.

rabernat:
> inferred from the grid variables that the user puts in

In general, I like the idea of inferring as much as possible from the inputs. However, it also means more work for gcm-filters to understand the meaning of all the different grid variables. So this question is closely related to the discussion in https://github.com/ocean-eddy-cpt/gcm-filters/pull/23#issuecomment-779341675.

Issue #6, opened 2021-01-11
Define kernels for filtering vectors


rabernat:
For models on curvilinear grids, u and v components get mixed together. So we will need to define kernels that effectively do viscosity (rather than diffusion).

rabernat:
Now that we have the implementation worked out in more detail, we can think about how to generalize our approach to vector laplacians.

The main difference between the laplacians that we already have and the vector laplacians is that the vector ones input and output two fields instead of one.

We could imagine defining kernels like
```python
class VectorLaplacian(BaseLaplacian):
    def __call__(self, field_u, self_field_v):
        # do stuff
        return u_component, v_component
```

We may also have to generalize the `Filter` class's `apply` method to account for the potentially variable number of arguments. Ideally the kernel should "know" about its expected number of inputs / ouputs, e.g. via a parameter like `n_inputs`.

iangrooms:
I have a theory question about the vector Laplacian: What should it conserve? For the scalar Laplacian we want to conserve the integral, but for the vector Laplacian I'm guessing it should be something like angular momentum (which is a vector).

rabernat:
Would be keen to get input from @StephenGriffies and @Hallberg-NOAA: they surely must think about this question in designing dissipation operators for GCMs. We basically just want to reproduce those as closely as possible in this package.

StephenGriffies:
When considering friction in the velocity equation, it is simplest to formulate the operator as the divergence of a symmetric and trace-free stress tensor.  Symmetry means the operator conserves angular momentum. Trace-free means the operator does not modify the mechanical pressure.  When the viscosity is formulated in a simple manner, then the divergence of the stress tensor can be reduced to a Laplacian operator. But in general it does not get that simple when the viscosities are spatially dependent.  

Apologies for the self-referencing, but I give a tutorial on these matters in Chapter 17 of my book.

NoraLoose:
@ElizabethYankovsky: The vector Laplacian in Steve's tutorial (Chapter 17) reminds me of the Laplacian form that you were originally thinking to implement. Sorry if I diverted you and @arthurBarthe away from implementing the vector Laplacian. When we talked I thought we could get away with the simpler scalar Laplacian, even for velocities. I was wrong... Did you further look into the vector Laplacian? It would be great if we could get this to work as a group!

ElizabethYankovsky:
[...]

PR #26, opened 2021-02-17
Laplacian on POP tripolar grid

NoraLoose:
This adds the new class `POPTripolarSimpleLaplacian` for a new `GridType`: `POP_TRIPOLAR_GRID`. This new kernel is identical to the `CartesianLaplacianWithLandMask` kernel, but --in addition-- handles the exchanges across the POP tripolar boundary condition correctly (so far: for tracer fields) by 
* grabbing the uppermost row, where convention is `(nlat, nlon)`
* mirroring the row
* appending it to the data as new uppermost row
* apply `CartesianLaplacianWithLandMask`-like operation
* disregard appended row,

where all operations above are compatible with dask and work on GPUs.

The `POPTripolarSimpleLaplacian` is useful for filtering with fixed coarsening factor (rather than with fixed length scale): it does not take any input grid variables other than `wet_mask`, because it assumes dx = dy = 1.

I am labeling this as "draft" because more things have to be done, for instance:
* Implement northern boundary exchanges for velocity fields (which is more complicated, see also the discussion [here](https://github.com/xgcm/xgcm/issues/194))
* Re-write this Laplacian so that it calls the superclass’s (`CartesianLaplacianWithLandMask`'s)` __init__() and __call__() method. (This should be easy to do.)

Despite not being finished, I am submitting this as a PR because it contains most of the pieces that we need for the gcm-filters-paper over [here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/4) - see [this](https://github.com/NoraLoose/gcm-filters/blob/POP-simple-Laplacian/docs/POP_tripolar_grid_Laplacian.ipynb) notebook for comparison of execution times.


rabernat:
This is great Nora! Thanks so much for your hard work on this.

I'm very interested in your benchmarking. It's great that the GPU version is so fast. But I'm also a bit concerned by how slow the CPU version is. Theoretically, we should be able to get within an order of magnitude of the fortran implementation.

However, I think our first priority here should be correctness, and then we can worry about optimization.

Let me know when you're ready for a review of this.

rabernat:
Follow up questions about your CPU-based timing. You said the Fortran implementation took 24 s vs. 6 min for the numpy CPU version. That's just over a factor of 10. Two questions:
- Were the data already in memory before doing this timing? Or are we including the time to load the data from disk in those numbers?
- How much parallelism was used in each case?
- Have you compared the serial execution time on a single level?

gcm_filters/kernels.py
```diff
@@ -91,6 +93,108 @@ def __call__(self, field: ArrayType):
...
+@dataclass
+class POPTripolarSimpleLaplacian(CartesianLaplacianWithLandMask):
+    """̵Laplacian for POP tripolar grid geometry with land mask, but assuming that dx = dy = 1
+
+    Attributes
+    ----------
+    wet_mask: Mask array, 1 for ocean, 0 for land
+    position: string
+        - 'T': Laplacian is applied for field defined at T-points
+        - 'U': Laplacian is applied for field defined at U-points
```
rabernat, line 105: Rather than having a "position" here, I imagined that these would actually be completely separate kernels, i.e.
- `POPTripolarSimpleLaplacianTpoint`
- `POPTripolarSimpleLaplacianUpoint`

NoraLoose: Thanks @rabernat. This was exactly a question I had for you. Having separate kernels will make things a lot easier in other parts of the code.

NoraLoose: I will try to finish up this PR this afternoon, and let you know when it's ready to review.

NoraLoose:
Coding up the Laplacian for fields that live on U-points turns out to be non-trivial. Here is a schematic that I made, which will hopefully help to review this PR:

[tripolarBgrid_exchanges.pdf](https://github.com/ocean-eddy-cpt/gcm-filters/files/6012540/tripolarBgrid_exchanges.pdf)

rabernat:
> Coding up the Laplacian for fields that live on U-points turns out to be non-trivial.

This reminds me of the hell we went through to implement exchanges in the MITgcm LLC! 😈  You can see some of that mess here: https://github.com/MITgcm/xmitgcm/blob/master/xmitgcm/llcreader/llcmodel.py

NoraLoose:
The northern boundary fold lies exactly on the northern edge of the uppermost T-cells. As a result, the northernmost U-points **lie on** the fold, and one of the U-points is exactly equal to the pivot point (see upper, middle schematic). 

Any POP U-field will satisfy the following property: The U-points right of the pivot points are a copy of the U-points left of the pivot point, but of opposite sign (see right, lower schematic). The same is true for the U wet mask, but here values are of course not inverted. Here is a plot that verifies what I said in this paragraph:
northern_velocity_tripolarBgrid

I have a question for people who worked with tripolar grids before, maybe @gustavo-marques or @iangrooms. How do you even define an integral over the global domain for U-fields? It seems that U-cells along the fold are "non-orientable" to say it in differential geometry lingo. Maybe I am just confused. I am asking because I don't think it makes sense to impose integral conservation of the Laplacian on U-fields (what does this even mean)?

NoraLoose:
> This reminds me of the hell we went through to implement exchanges in the MITgcm LLC! 😈 You can see some of that mess here: https://github.com/MITgcm/xmitgcm/blob/master/xmitgcm/llcreader/llcmodel.py

I am very familiar with the pain that the MITgcm LLC grid entails. :smile: And with the llcreader, too - it's such an awesome tool!

iangrooms:
> I have a question for people who worked with tripolar grids before, maybe @gustavo-marques or @iangrooms. How do you even define an integral over the global domain for U-fields?

I haven't dealt with this problem before; my only experience is with tracer cells. The raw data is not meridonal/zonal, it is along-grid. So the "grid-north" direction changes depending on which side of the fold you're on, hence the sign flip. I conjecture that if you do the following there won't be a problem:
1. Only filter zonal/meridional velocities, not grid-aligned velocities. This should fix the orientability problem, and lead to velocities on the fold that are just duplicated, rather than duplicated with a sign change.
2. To define the global integral on U cells, only use half the cells on the fold (since they're all duplicated). 

I might be completely off-base here - this is just an idea.

NoraLoose:
> Only filter zonal/meridional velocities, not grid-aligned velocities. This should fix the orientability problem, and lead to velocities on the fold that are just duplicated, rather than duplicated with a sign change.

Are you suggesting that our filter code should check whether the user has put in a field where values on the fold are duplicated, rather than duplicated & inverted?

> To define the global integral on U cells, only use half the cells on the fold (since they're all duplicated).

So which side of the fold do you pick? Your integral will not be the same in these two cases.

iangrooms:
> Are you suggesting that our filter code should check whether the user has put in a field where values on the fold are duplicated, rather than duplicated & inverted?

I wasn't thinking about it in terms of the software package, but I guess that makes sense. 

> So which side of the fold do you pick? Your integral will not be the same in these two cases. I guess your point 2. requires point 1, but this is maybe what you mean.

Yes, part 2 presumes part 1, otherwise you get different answers depending on which half of the fold you use in the integral

NoraLoose:
> Only filter zonal/meridional velocities, not grid-aligned velocities. This should fix the orientability problem, and lead to velocities on the fold that are just duplicated, rather than duplicated with a sign change.

If we do this, then any current through the north pole (e.g., the Transpolar Drift-ish) will also be filtered in a weird way: Meridional velocities will be inverted at the north pole, so the filtered field will be approximately zero there?

I guess the same problem would arise for the grid-aligned velocities...

Murray_1996

NoraLoose:
Or maybe @arthurBarthe, @ElizabethYankovsky, @LaureZanna have experience with filtering U fields on a tripolar B-grid and the issues mentioned above?

iangrooms:
Good point. We might need the vector Laplacian for velocity fields at the pole. I am curious how viscosity is implemented in models with the tripole seam though.

Hallberg-NOAA:
Hi,

  We always use locally orthogonal grids, and we keep the velocity
components that are aligned with the grids. This has the nice feature that
when we apply a Laplacian, we do not need to rotate the velocity
components, and all of the grid metric terms are accounted for very neatly
within the operators themselves, without the need to add any extra terms,
to 2nd order accuracy.  They also work very neatly with constructing the
tensor components, especially with a C-grid discretization, so you
automatically get certain conservation and rotational invariance
properties.  I think that we might have described this in appendix A of
Griffies and Hallberg, MWR, 2000.

  - Bob Hallberg

 Dr. Robert Hallberg
 Oceanographer
 NOAA Geophysical Fluid Dynamics Laboratory

NOAA GFDL                             || Phone: (609) 452-6508
Princeton University Forrestal Campus || Cell:  (732) 599-0459
201 Forrestal Road                    || Fax:   (609) 987-5063
Princeton, New Jersey 08540-6649      || Email: Robert.Hallberg@noaa.gov

iangrooms:
Hi Bob,
     Thanks! We're specifically interested in how this works across the tripole seam. I imagine that no "tiles" cross the seam, so the question is about how the code pulls in the "halo" for velocities from across the other side of the tripole seam. It seems straightforward for tracer-cell variables, but we're having trouble figuring it out for velocities.
   Best,

Ian


Ian Grooms
Assistant Professor, Applied Mathematics
University of Colorado, 526 UCB
Boulder, CO 80309-0526
Office: 320 ECOT (Engineering Center)
https://www.colorado.edu/amath/grooms


________________________________

Hallberg-NOAA:
Hi,

  The only thing that we have to do that is special at the tripolar fold is
that the halo update for the vector components also includes a rotation, so
that the halo points have the negative of the velocity components on the
counterparts.  We always do a halo update, even if we are running on a
single processor.

  - Bob Hallberg

 Dr. Robert Hallberg
 Oceanographer
 NOAA Geophysical Fluid Dynamics Laboratory

NOAA GFDL                             || Phone: (609) 452-6508
Princeton University Forrestal Campus || Cell:  (732) 599-0459
201 Forrestal Road                    || Fax:   (609) 987-5063
Princeton, New Jersey 08540-6649      || Email: Robert.Hallberg@noaa.gov

iangrooms:
Based on Bob's comments, I think that we _could_ implement the tracer Laplacian for velocity input by doing a halo with rotation. Whereas tracer was something like `halo[i] = last_row[Nx-i]` the velocity should something like `halo[i] = -penultimate_row[Nx-i]`. The switch from `last_row` to `penultimate_row` is because of the staggering of the variables. 

But it also sounds like this is a bad idea. (It was not Bob's idea, so this is not a criticism.) The integral of a tracer is a natural quantity to conserve, but the integral of a component of velocity projected into some local coordinate system does not seem like a natural quantity to conserve. Sounds like we should consider a vector Laplacian that conserves some kind of momentum or angular momentum integral.

NoraLoose:
> Based on Bob's comments, I think that we could implement the tracer Laplacian for velocity input by doing a halo with rotation. Whereas tracer was something like halo[i] = last_row[Nx-i] the velocity should something like halo[i] = -penultimate_row[Nx-i].

@iangrooms, this is exactly what I did, and what I was trying to visualize in the colorful schematic above, lower right panel. See also the newly commited code in `POPTripolarSimpleLaplacianUpoint` in `kernels.py`:
```
        # prepare data for northern boundary exchanges
        folded = data[..., [-2], :]  # grab second northernmost row
        folded = -folded[..., ::-1]  # mirror and invert it
        folded = np.roll(folded, -1, axis=-1)  # shift by 1 cell to the left

        data = np.concatenate((data, folded), axis=-2)  # append mirrored row
```

rabernat:
> But it also sounds like this is a bad idea. (It was not Bob's idea, so this is not a criticism.) The integral of a tracer is a natural quantity to conserve, but the integral of a component of velocity projected into some local coordinate system does not seem like a natural quantity to conserve. Sounds like we should consider a vector Laplacian that conserves some kind of momentum or angular momentum integral.

It sounds like we have to decide whether to move forward with the approach Nora has implemented here or tackle the more difficult problem of vector laplacians (see #6). My personal view is that we should just bite the bullet and do the vector laplacian, as this is the most correct way to handle velocities on general curvilinear grids, including POP tripolar. However, perhaps we want to merge this PR as is and then handle that in a future PR? Thoughts?

iangrooms:
I hesitate to merge this PR as is because it might lead to users unknowingly using code that shouldn't be used. One option is to treat this as a scalar Laplacian on the U points, which I think only requires changing a sign in the current code (i.e. not reflecting the values pulled in across the tripole seam). This will yield garbage results when applied to velocity components within one filter radius of the tripole seam, but in a sense that's OK because this probably shouldn't be used for velocity components anyways. Previously I thought we might get away with scalar Laplacian but I now realize that was wrong. (You can still use the scalar Laplacian to filter velocity over a region away from the Arctic; it's not ideal, but it's a stopgap.) I agree that we need to get the vector Laplacian going for velocity fields.

NoraLoose:
I agree with you, Ryan and Ian. The exercise of implementing the U-point Laplacian on the tripolor grid really convinced me that we have to tackle the vector Laplacian issue. Maybe it is best to not provide a scalar POP tripolar Laplacian on U-points at all, because this may only lead to errors and confusion among users. I think all diagnostic U-point variables from POP are actually vector components, so we don't need the scalar U-point Laplacian. I will remove it from this PR.

rabernat:
My understanding is that #29 supersedes this PR, correct? If so, should we close this?

NoraLoose:
> My understanding is that #29 supersedes this PR, correct? If so, should we close this?

Correct. I'm closing this PR.

PR #28, opened 2021-02-22
Option to apply anisotropic diffusion in IrregularLaplacianWithLandMask

gustavo-marques:
This PR adds two additional arguments (kappa_s and kappa_w) to the IrregularCartesianLaplacianWithLandMask kernel, enabling one to apply an anisotropic diffusion with this type of grid. Right now both arguments are required, but ideally, we should set them to one by default and make them optional. I could not figure out a way for doing that without modifying the function `required_grid_args`.


gcm_filters/kernels.py
```diff
@@ -113,6 +115,8 @@ class IrregularCartesianLaplacianWithLandMask(BaseLaplacian):
     dxs: ArrayType
     dys: ArrayType
     area: ArrayType
+    kappa_w: ArrayType
+    kappa_s: ArrayType
```
NoraLoose, line 119: Thanks for this addition! It would be great to include a check that the user does not input kappa > 1, which would cause the filter to blow up (as discussed by @iangrooms and @sdbachman on slack).

gcm_filters/kernels.py
```diff
@@ -132,8 +136,8 @@ def __call__(self, field: ArrayType):
             (out - np.roll(out, -1, axis=-2)) / self.dys * self.dxs
         )  # flux across southern cell edge
 
-        wflux = wflux * self.w_wet_mask  # no-flux boundary condition
-        sflux = sflux * self.s_wet_mask  # no-flux boundary condition
+        wflux = self.kappa_w * wflux * self.w_wet_mask  # no-flux boundary condition
+        sflux = self.kappa_s * sflux * self.s_wet_mask  # no-flux boundary condition
```
NoraLoose, line 140: I think we could make the filter faster if we multiply `self.w_wet_mask` and `self.s_wet_mask` by their respective kappas in the post_init method, to avoid re-computation in every Laplacian iteration. What do you think?

gcm_filters/kernels.py
gustavo-marques: Ok, I will add this check.

gcm_filters/kernels.py
gustavo-marques: Yes, I agree. I talked about doing this during our last meeting, but I completely forgot to change that. I will make this change.

gustavo-marques:
@NoraLoose, I made the modifications requested. Please let me know if you have any other suggestions.

gcm_filters/kernels.py
```diff
@@ -121,8 +121,20 @@ class IrregularCartesianLaplacianWithLandMask(BaseLaplacian):
...
-        self.s_wet_mask = self.wet_mask * np.roll(self.wet_mask, -1, axis=-2)
+        err_w = np.where(self.kappa_w > 1.)[0]
+        if err_w.size > 0 :
+            raise ValueError(f"There are kappa_w values > 1 and this can cause the filter to blow up."
+                             f"Please make sure all kappa_w are <=1.")
+
+        err_s = np.where(self.kappa_s > 1.)[0]
+        if err_s.size > 0:
+            raise ValueError(f"There are kappa_s values > 1 and this can cause the filter to blow up."
+                             f"Please make sure all kappa_s are <=1.")
```
NoraLoose, line 132: Thanks for adding this! 

It would be nice to "test this test"; that is, we could add a test to `test_kernels.py` which checks that the error gets actually triggered, if the user puts in `kappa > 1`. I imagine the test could look almost identical to the test I wrote here (in my test I'm checking that an error gets triggered for the `TRIPOLAR` Laplacians if the wet mask has unexpected values):

https://github.com/NoraLoose/gcm-filters/blob/0591b7aae5045e1f90363d5229dec68563b7e0ac/tests/test_kernels.py#L72-L84

Does this make sense?

rabernat: I know I am late to this PR, but I will weigh in quickly with my opinion.

I don't think we should do these kinds of checks. They may slow things down considerably, especially for large arrays. And where do we stop? Should we also check `area`?

 Instead, we should document what's required (positive values) and trust the user to read it.

NoraLoose: I suggested this because the check would be part of the \__post_init\__ method. Since the \__post_init\__ method is ideally only called once (I know the realization of this is still on our todo list) and the check acts on one 2D array only, I was hoping it wouldn't slow things down by very much. But I'm okay with not doing these kind of checks, and instead documenting everything carefully.

rabernat: > Since the __post_init__ method is ideally only called once (I know the realization of this is still on our todo list)

Yes that would be ideal! But for now that it not how it works...it looks like `__post_init__` is unfortunately getting called on every chunk. 😢

gustavo-marques: @rabernat and @NoraLoose, Thanks for the feedback. I will convert this PR to draft until we decide on the right approach.

rabernat: I have done some profiling and convinced myself that the `__post_init__` cost is not significant in `IrregularCartesianLaplacianWithLandMask`. So I have no problem with these checks.

Later down the road we can do some more serious profiling and try to optimize performance further. But doing that now is premature.

Sorry for slowing things down. @gustavo-marques you can mark as ready for review again and I'll merge.

gustavo-marques: There is already a simple test for this in `test_filter.py`.  Should I add something else?

rabernat: Yes, as Nora said above, you need to test the if block you have introduced. Currently the code that checks for the kappas < 1 is untested. You should use `pytest.raises` (as in https://github.com/NoraLoose/gcm-filters/blob/0591b7aae5045e1f90363d5229dec68563b7e0ac/tests/test_kernels.py#L83-L84) to trigger the error in the test function.

gcm_filters/kernels.py
```diff
@@ -113,12 +115,26 @@ class IrregularCartesianLaplacianWithLandMask(BaseLaplacian):
...
+    kappa_w: ArrayType
+    kappa_s: ArrayType
 
     def __post_init__(self):
         np = get_array_module(self.wet_mask)
 
-        self.w_wet_mask = self.wet_mask * np.roll(self.wet_mask, -1, axis=-1)
-        self.s_wet_mask = self.wet_mask * np.roll(self.wet_mask, -1, axis=-2)
+        err_w = np.where(self.kappa_w > 1.)[0]
+        if err_w.size > 0 :
```
rabernat, line 125:
```suggestion
        if np.any(self.kappa_w > 1) :
```

gustavo-marques:
@rabernat, thanks for the suggestion and clarification.  It should be good to go now.

iangrooms:
The check for whether there are any `kappa>1.0` is great. I suggest also checking that `kappa=1` at at least one point, otherwise the filter's scale will not be equal to `filter_scale` anywhere in the domain.

gustavo-marques:
Got it, sorry I miss understood that. I will make this change and update the PR.

gustavo-marques:
If we do not find a solution for making kappa_s and kappa_w optional arguments, I will update tutorial_irregular_grid.ipynb.

jbusecke:
I have to admit, that I am not really deep in the code base so far, but [this](https://docs.python.org/3/library/dataclasses.html#module-level-decorators-classes-and-functions) seems relevant:

```
@dataclass
class C:
    a: int       # 'a' has no default value
    b: int = 0   # assign a default value for 'b'
```
I am however not sure how to deal with an array. Perhaps setting the default to `None` and then dealing with 'filling' these values in the `__post_init__` method? Really just a guess though.

gustavo-marques:
The following modification to [required_grid_args](https://github.com/ocean-eddy-cpt/gcm-filters/blob/master/gcm_filters/kernels.py#L46-L50) is a possible solution. Toughts? 

```
    def required_grid_args(self):
        try:
           tmp_list = list(self.__annotations__)
           if 'kappa_s' in tmp_list:
              tmp_list.remove('kappa_s')
           if 'kappa_w' in tmp_list:
              tmp_list.remove('kappa_w')
           return tmp_list
        except AttributeError:
           return []
```

NoraLoose:
I think we got stuck with this PR because of this issue:

> It would be useful if we could make both kappa_s and kappa_w optional arguments with the default values = 1.

Since introducing optional grid variables seems to collide with `required_grid_args`, I suggest to make `kappa_s` and `kappa_w` required variables for the time being, so we can move forward with this PR. 

I think `kappa_s` and `kappa_w` got sneaked in via different PRs already but the tests for the kappas are not yet in the master branch and would be great to have. 

@gustavo-marques, let me know if this sounds good to you. Thanks!

gustavo-marques:
@NoraLoose, sounds good.  This branch has conflicts now. Are you going to handle that or do you want me to do it?

NoraLoose:
@gustavo-marques if you could try to resolve the conflicts, that would be much appreciated. Thanks!

PR #29, opened 2021-02-24
T-Laplacian for POP tripolar grid for varying dx, dy

NoraLoose:
This adds the `POP_TRIPOLAR_T_GRID` and the associated Laplacian `POPTripolarLaplacianTpoint`. This addition is for creating general filters on the POP tripolar grid and needs the user to input stuff like `dxe`, `dye`, `dxn`, `dyn` - grid length spacing of the eastern and northern T-cell edges. (In contrast, the Laplacian in PR #26 is only for fixed factor filtering and doesn't need any grid information except `wet_mask`.)

I also included a notebook that compares the action of the 5 different Laplacian kernels that we have so far (on POP model output), and compares the timing of these on CPU.

This PR sits on top of #26, so #26 should be merged first.


rabernat:
Is this ready for review?

NoraLoose:
Yes, ready for review.

gcm_filters/kernels.py
```diff
@@ -12,12 +12,29 @@
 
 # not married to the term "Cartesian"
 GridType = enum.Enum(
-    "GridType", ["CARTESIAN", "CARTESIAN_WITH_LAND", "IRREGULAR_CARTESIAN_WITH_LAND"]
+    "GridType",
+    [
+        "CARTESIAN",
+        "CARTESIAN_WITH_LAND",
+        "IRREGULAR_CARTESIAN_WITH_LAND",
+        "POP_SIMPLE_TRIPOLAR_T_GRID",
```
rabernat, line 20: This kernel strikes me as very similar to the `CARTESIAN_WITH_LAND`, the only difference being the boundary condition. Do we want to think about harmonizing the names? What about `TRIPOLAR_CARTESIAN_WITH_LAND`?

Or maybe we want to rename things like `REGULAR`, `REGULAR_WITH_LAND`, `TRIPOLAR_REGULAR_WITH_LAND`, etc?

rabernat (commented):
This is really excellent work Nora! I tried hard to find areas to improve your code but could not! 😊  I especially like your tests for the tripolar boundary condition.

My only question is about our terminology to describe these different grids.

gcm_filters/kernels.py
NoraLoose: Thanks, Ryan! Good point about harmonizing the names. I like your suggestion, and vote for: `REGULAR`, `REGULAR_WITH_LAND`, `IRREGULAR_WITH_LAND`, `TRIPOLAR_REGULAR_WITH_LAND`, `TRIPOLAR_IRREGULAR_WITH_LAND`. Or we could keep the latter as `POP_TRIPOLAR_T_GRID`, and change it later to something more generic (e.g., `TRIPOLAR_IRREGULAR_WITH_LAND`), as you suggested here: https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/18#issuecomment-790781694

rabernat: I think we should some distinguish between three types of grids:

- Regular + Rectangular: dx=dy=const, dimensions are globally orthogonal (what VTK calls "image data")
- Irregular + Rectangular, dimensions are globally orthogonal, but dx and dy can vary independently (what VTK calls "rectilinear"
- Curvilinear, dimensions are locally orthogonal, but the grid curves (what vtk calls "structure")
- Unstructured mesh (e.g. triangles, hexagons, etc.) - probably out of scope for now

In addition to this, we seem to have two independent option spaces
- Boundary conditions (e.g. singly or doubly periodic, tripolar, cubed-sphere, etc.)
- Land mask or not

(VTK reference: http://www.bu.edu/tech/support/research/training-consulting/online-tutorials/vtk/)

We don't need to resolve all of this now, but maybe we can use these concepts to inform the naming convention.

Down the road (future PR), we could maybe even use multiple inheritance to define kernels, like

```python
class POPCurvilinearGrid(CurvilinearGrid, GridWithLandMask, TripolarBoundaryCondition):
    pass
```

NoraLoose: Agreed, these are useful guidelines. 

The thing is that all Laplacians that we have implemented so far can be used on curvilinear grids, even the ones where dx=dy=1, e.g., `CARTESIAN_WITH_LAND` or `POP_SIMPLE_TRIPOLAR_T_GRID`. The latter are the Laplacians for fixed factor filtering, but they work for the POP **curvilinear** grid, for instance.

I like your idea about multiple inheritance to define kernels - this is more or less what I have imagined too. But I appreciate that this is maybe more of a long-term goal, and will come with its own challenges.

NoraLoose: Since all Laplacians are good for curvilinear grids, I decided to go with 
* "regular": Laplacian needs **no** grid info (for fixed factor filtering)
* "irregular": Laplacian needs grid info (for more general filtering, e.g., with fixed scale)

I'm happy to change this if there are better suggestions, either now or later.

gcm_filters/kernels.py
```diff
@@ -141,9 +158,122 @@ def __call__(self, field: ArrayType):
...
+    """
+
+    wet_mask: ArrayType
+
+    def __post_init__(self):
+        np = get_array_module(self.wet_mask)
+
+        # check that southernmost row of wet mask has only zeros
+        if self.wet_mask[..., 0, :].any():
+            raise AssertionError("Wet mask requires zeros in southernmost row")
```
NoraLoose, line 180: @rabernat, related to your comment on checks [here](https://github.com/ocean-eddy-cpt/gcm-filters/pull/28#discussion_r594690734), I'm assuming you would suggest to remove this check (and associated test) as well. I added this check because this is a global Laplacian (on a tripole grid), but the southern boundary must be land in order to disable the periodic boundary condition in y-direction. Otherwise, the filtered values in the very south will be contaminated by stuff in the Arctic. 

Again, an alternative is to document the requirement of wet mask zeros in the southernmost row carefully, and I am happy to remove this check.

rabernat: Good point.

I will think about this and get back to you. Perhaps fixing the problem at the root (only call `__post_init__` once) would be the best path forward.

rabernat: I did some profiling and decided that this is not actually a big performance hit. So we can leave things as are for now.

gcm_filters/kernels.py
```diff
@@ -12,12 +12,29 @@
...
 ALL_KERNELS = {}  # type: Dict[GridType, Any]
 
 
+def _prepare_tripolar_exchanges(field):
+    """Auxiliary function that prepares T-field for northern boundary exchanges on tripolar grid"""
+    np = get_array_module(field)
+
+    folded = field[..., [-1], :]  # grab northernmost row
+    folded = folded[..., ::-1]  # mirror it
+    field_extended = np.concatenate((field, folded), axis=-2)  # append it
```
rabernat, line 34: 👌 perfect style 🤩

rabernat (approved):
This is great work. I think we can merge it.

NoraLoose:
Good that we hadn't merged this. I discovered a sign error in my rolling operations (which I made so consistently throughout my code that the tests did not uncover it). I fixed it, and we should now be ready to merge.

rabernat:
Nice!

Can you think what sort of test would have caught your sign error? Is there a new type of test we should consider adding to the kernel test suite?

rabernat:
> Is there a new type of test we should consider adding to the kernel test suite?

Hi Nora! Any chance we can think of a test that would have caught the sign error? If not, no big deal.

NoraLoose:
Yes, I think I have an idea. I have time to work on this today. 

If I come up with a successful test, would you like me to submit it as part of this PR? (I would design the test for all irregular Laplacians, i.e., all Laplacians that need grid info `dx`, `dy`.)

NoraLoose:
I added two new tests in `test_kernels.py` that would have caught the sign error: `test_use_correct_dy`, `test_use_correct_dx`. I verified that the old version of the code (before commit `ba228f7`) would have failed the test.

tests/test_kernels.py
```diff
@@ -26,18 +27,193 @@ def grid_type_field_and_extra_kwargs(request):
...
+            diffused[random_yloc, random_xloc - 1],
+            diffused[random_yloc, random_xloc + 1],
+            atol=1e-12,
+        )
+
+
+################## Tripolar grid tests ##############################################
+tripolar_grids = [
+    member
+    for name, member in GridType.__members__.items()
```
rabernat, line 181: Why do you need to use `GridType.__members__.items()` here? Does `for name in GridType` not work?

tests/test_kernels.py
```diff
@@ -26,18 +27,193 @@ def grid_type_field_and_extra_kwargs(request):
...
+
+
+################## Irregular grid tests ##############################################
+# Irregular grids are grids that allow spatially varying dx, dy
+
+# The following definition of irregular_grids is hard coded; maybe a better definition
+# would be: all grids that have len(required_grid_vars)>1 (more than just a wet_mask)
+irregular_grids = [
+    member
+    for name, member in GridType.__members__.items()
```
rabernat, line 71: Same

rabernat (changes requested):
Thanks again @NoraLoose for this amazing PR! And sorry for being slow to review.

My only real concern that, if we are changing the names of things, our documentation tutorials need to be updated for consistency. In https://gcm-filters.readthedocs.io/en/latest/tutorial.html#create-a-filter for example we still use `CARTESIAN_WITH_LAND`, etc.

Also a few nit-picking code style issues.

With this we will finally be ready to merge. Thanks for your patience and your excellent work.

tests/test_kernels.py
NoraLoose: `for name in GridType` raises the error:
```
TypeError: argument of type 'GridType' is not iterable
```

NoraLoose:
Thanks for the review @rabernat! 

* I will add some comments on the sign convention of the `shift` argument, as you suggest. 
* Re the Laplacian name changes: I'm happy to update the tutorials. Do we think that the new naming convention is superior to the old one? It does shorten the names by a little bit...

tests/test_kernels.py
rabernat: That is puzzling to me. `GridType` is an enum. The following works.

```
import enum
GridType = enum.Enum("GRID_TYPE", ["A", "B"])
assert [gt.name for gt in GridType] == ["A", 'B']
```

NoraLoose:
I think I have made all changes you suggested. Please let me know if you can think of anything else. Thanks!

rabernat:
Fantastic! Thanks again for your excellent work on this. I'm excited to keep pushing it forward!

PR #34, opened 2021-04-23
Vector Laplacian on C-grid

NoraLoose:
### Summary
This PR adds a vector Laplacian on a C-grid mimicking the [MOM6 implementation of viscosity](https://github.com/NOAA-GFDL/MOM6/blob/dev/gfdl/src/parameterizations/lateral/MOM_hor_visc.F90), based on the friction operator suggested by Griffies & Hallberg, 2000. 

### Tests
While our scalar Laplacians are tested to conserve the area integral, the new vector Laplacian is designed to **preserve angular momentum**. A corollary of angular momentum conservation is that the vector Laplacian is invariant under solid body rotations. The latter is the test that I implemented for the vector Laplacian. Shoutout to @adcroft, who helped with designing this vector Laplacian test!

### Numerical instability
While the viscosity-based filter (i.e., the filter using the vector Laplacian) seems to produce reasonable filtered fields (see also figure below), it unfortunately seems way more sensitive to numerical instability than the diffusion-based filter (the filter using the scalar Laplacian). Tests with NeverWorld2 data suggest that the old stage order (before #32) is more stable for the vector Laplacian. I don't know why... I am happy to revert this PR to "draft" until we have figured this out.

### A few other (minor) notes:
* For now, `filter.py` has two separate functions:`_create_filter_func` and `_create_filter_func_vec`; as well as methods: `apply` and `apply_to_vector`, to treat diffusion- vs. viscosity-based filters. There is probably a more elegant way to do this. Suggestions are welcome!
* Naming convention of grid variables follows MOM6.
* I implemented anisotropy aligned with the *grid directions*. Work left for a future PR: A few minor additions would allow anisotropy in arbitrary direction (aligned with spatially varying normal `n=(n_1, n_2)` that is input by the user) a la Smith and McWilliams, 2003 - see also MOM6 code and [this issue](https://github.com/NOAA-GFDL/MOM6/issues/1368) on how to correctly implement this.


NoraLoose:
fixed_factor_filters_visc_u_deg16_fac16_layer0
fixed_factor_filters_visc_v_deg16_fac16_layer0_log

The figures above compare diffusion-filtered vs. viscosity-filtered velocity for NeverWorld2 data. The fields are filtered from 1/16 degree to 1 degree.

iangrooms:
I discussed this with @NoraLoose yesterday and she suggested that the numerical stability could be related to the fact that the lat/lon Neverworld2 grid is extremely anisotropic near the poleward boundaries. I support merging this PR and then continuing to work on numerical stability in the future. It would be nice to see if the stability problems persist on a more realistic (e.g. Mercator) grid, with cells that are closer to unit aspect ratio.

NoraLoose:
> the numerical stability could be related to the fact that the lat/lon Neverworld2 grid is extremely anisotropic near the poleward boundaries

On the other hand, the diffusion-based filter also knows about the extreme anisotropy of the NeverWorld2 grid. Nonetheless, the diffusion-based filter works okay.

Another possibility is the handling of boundaries: this is where the diffusion-based and viscosity-based filter differ quite a bit. You can even see that in the figures above (right-most panels). There seems to be some noise building up, particularly at the southern and northern boundaries. One could potentially look into whether there is a better way to handle the wet masks than done here:

https://github.com/NoraLoose/gcm-filters/blob/7f9c8bc81ef17a291e6fdc091b6dc9906bcc9b67/gcm_filters/kernels.py#L342-L345

rabernat:
Sorry for taking so long to respond to this. I was on vacation last week and have still not managed to catch up! 🙃 

I have just created the group @ocean-eddy-cpt/gcm-filters-dev which now has admin rights on this repo. It includes myself, @NoraLoose and @iangrooms, but we can add anyone else interested in maintaining.

At this point, I don't want to be a bottleneck on progress here. As long as someone from @ocean-eddy-cpt/gcm-filters-dev reviews and approves the PR, they should feel empowered to merge.

I can try to review early next week if desired.

gcm_filters/filter.py
```diff
@@ -83,6 +83,7 @@ def _compute_filter_spec(
             else:  # ndim==2
                 n_steps = np.ceil(6.4 * filter_scale / dx_min).astype(int)
 
+    print("n_steps = %i" % n_steps)
```
rabernat, line 86: No print statements in the code.

tests/test_filter.py
```diff
@@ -147,3 +234,34 @@ def test_filter(grid_type_and_input_ds, filter_args):
...
+
+    # check conservation under solid body rotation: u = cos(lat), v=0;
+    # this test fails due to numerical instabilities
+    # data_u = np.cos(geolat_u/360*2*np.pi)
+    # data_v = np.zeros_like(data_u)
+    # da_u = xr.DataArray(data_u, dims=["y", "x"])
+    # da_v = xr.DataArray(data_v, dims=["y", "x"])
+    # filtered_u, filtered_v = filter.apply_to_vector(da_u, da_v, dims=["y", "x"])
+    # xr.testing.assert_allclose(filtered_u, 0.0, atol=1e-12)
+    # xr.testing.assert_allclose(filtered_v, 0.0, atol=1e-12)
```
rabernat, line 259: Why commented-out code? Probably best to avoid this.

rabernat (commented):
Some superficial comments.

gcm_filters/filter.py
NoraLoose: Thanks for catching this! The print statement is now removed.

tests/test_filter.py
NoraLoose: I fixed the test and uncommented it. 

(I had wrongly assumed that numerical instability had caused the test to fail, but it was actually a small bug in the test - which is now fixed.)

NoraLoose:
I made a small modification to the code. The previous version did a division by zero if the user provides `area_u`, `area_v` that contain zeros (as is the case for e.g., MOM6 output over land). This is now fixed via a conditional statement.
```
self.recip_area_u = np.where(self.area_u > 0, 1 / self.area_u, 0)
```
The code still raises a `RuntimeWarning: divide by zero encountered in true_divide` if the user runs the vector Laplacian, but only on CPUs (using the above line with `np.where`) - not on GPUs (using the above line with `cp.where`). Weird, and not optimal. 

The following version would avoid the warning on CPUs, but does not work on GPUs:
```
self.recip_area_u = np.divide(1, self.area_u, out=np.zeros_like(self.area_u), where=self.area_u!=0)
```

Can anyone think of a practical solution here? Maybe @jbusecke?

I also added a new test `test_zero_area` in `test_kernels.py` which reproduces the current warning (but checks that the division by zero does not actually happen in the Laplacian).

rabernat:
Good catch! Ready for another review here?

I don't know what's up with the RTD build 🤔

NoraLoose:
I'm also not sure why the Read the Docs build suddenly fails. In my last three commits I haven't changed anything that should affect the docs. Any ideas?

rabernat:
> Any ideas?

The most likely reason is the one of the packages that the docs depend on has released a new version with an incompatibility.

The quick solution is to figure out which one it is and then pin a specific version number here:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/64b3b8ecb944a6c290c743b14a4cc521ad3278c4/docs/environment.yml#L2-L22

Unfortunately the error does not immediately make it obvious which one. Myst-nb had a minor release (0.12.3) about a month ago, so that could be it: https://github.com/executablebooks/MyST-NB/releases/
You could try pinning myst_nb==0.12.2

NoraLoose:
> Good catch! Ready for another review here?

Sure, thanks! 👍

gcm_filters/kernels.py
```diff
@@ -290,6 +299,115 @@ def __call__(self, field: ArrayType):
 ALL_KERNELS[GridType.TRIPOLAR_POP_WITH_LAND] = POPTripolarLaplacianTpoint
 
 
+@dataclass
+class VectorLaplacian(BaseLaplacian):
```
rabernat, line 303:
```suggestion
class CgridVectorLaplacian(BaseLaplacian):
```

Maybe this name would be a bit more specific?

I also wonder whether we should use different base classes, e.g. `BaseScalarLaplacian`, `BaseVectorLaplacian`. This would clarify a few ambiguities in the code.

gcm_filters/kernels.py
```diff
@@ -290,6 +299,115 @@ def __call__(self, field: ArrayType):
...
+        self.dx2h = self.dxT * self.dxT
+        self.dy2h = self.dyT * self.dyT
+        self.dx2q = self.dxBu * self.dxBu
+        self.dy2q = self.dyBu * self.dyBu
+
+        # compute reciprocal of areas while avoiding division by zero
+        self.recip_area_u = np.where(self.area_u > 0, 1 / self.area_u, 0)
+        self.recip_area_v = np.where(self.area_v > 0, 1 / self.area_v, 0)
+
+    def __call__(self, ufield: ArrayType, vfield: ArrayType):
```
rabernat, line 356: My concern here is the the signature of this Laplacian does not match the signature of `BaseLaplacian`, which is

```python
def __call__(self, field):
```

I'm surprised that mypy did not complain about that. In any case, it supports the idea suggested above that we should have a separate base class for vector laplacians.

tests/test_filter.py
```diff
@@ -67,7 +67,11 @@ def test_filter_spec(filter_args, expected_filter_spec):
     # TODO: check other properties of filter_spec?
 
 
-@pytest.fixture(scope="module", params=list(GridType))
+vector_grids = [gt for gt in GridType if gt.name.startswith("VECTOR")]
```
rabernat, line 70: Parsing the name string does not seem like a robust way to detect vector laplacians. Instead we should somehow refactor our base data structures to store this information. In the meantime, simply hard-coding the grid types here would not be a bad way to go. There are not so many.

gcm_filters/filter.py
```diff
@@ -191,6 +191,53 @@ def filter_func(field, *args):
...
+    and whose subsequent arguments are the require grid variables
+    """
+
+    def filter_func_vec(ufield, vfield, *args):
+        # these next steps are a kind of hack we have to turn keyword arugments into regular arguments
+        # the reason for doing this is that Xarray's apply_ufunc machinery works a lot better
+        # with regular arguments
+        assert len(args) == len(Laplacian.required_grid_args())
+        grid_vars = {k: v for k, v in zip(Laplacian.required_grid_args(), args)}
+        laplacian = Laplacian(**grid_vars)
```
rabernat, line 208: What happens if the user passes a non-vector Laplacian here? It should not be allowed.

gcm_filters/filter.py
```diff
@@ -271,3 +318,22 @@ def apply(self, field, dims):
             dask="parallelized",
         )
         return field_smooth
+
+    def apply_to_vector(self, ufield, vfield, dims):
+        """Filter a vector field across the dimensions specified by dims."""
+
+        filter_func_vec = _create_filter_func_vec(self.filter_spec, self.Laplacian)
```
rabernat, line 325: We should be able to raise an informative error if a user tries to create to do a vector filter with a scalar laplacian I think we can resolve this by implementing the suggestion to have a separate `BaseScalarLaplacian` and `BaseVectorLaplacian` and then checking the type in the `Filter.__post_init__` method.

rabernat (commented):
This is awesome Nora. It will be so extremely useful in many different scenarios.

My main concern is that we are overloading the `BaseLaplacian` class a little too heavily. See detailed comments below.

gcm_filters/kernels.py
NoraLoose: Good point. Definitely a good idea to have two separate base classes.

gcm_filters/filter.py
NoraLoose: Here is an attempt to raise an error based on the two base classes:
https://github.com/NoraLoose/gcm-filters/blob/59bb142a8c01224138d00026000a1732873271de/gcm_filters/filter.py#L330-L334

Does this look good to you?

gcm_filters/filter.py
```diff
@@ -255,7 +303,12 @@ def __post_init__(self):
         self.grid_ds = xr.Dataset({name: da for name, da in self.grid_vars.items()})
 
     def apply(self, field, dims):
-        """Filter a field across the dimensions specified by dims."""
+        """Filter a field with scalar Laplacian across the dimensions specified by dims."""
+        if not issubclass(self.Laplacian, BaseScalarLaplacian):
+            raise ValueError(
+                f"Provided Laplacian {self.Laplacian} is a vector Laplacian. "
+                f"The .apply method is only suitable for scalar Laplacians."
```
rabernat, line 310:
```suggestion
                f"The ``.apply`` method is only suitable for scalar Laplacians."
```

gcm_filters/filter.py
```diff
@@ -271,3 +324,27 @@ def apply(self, field, dims):
             dask="parallelized",
         )
         return field_smooth
+
+    def apply_to_vector(self, ufield, vfield, dims):
+        """Filter a vector field with vector Laplacian across the dimensions specified by dims."""
+        if not issubclass(self.Laplacian, BaseVectorLaplacian):
+            raise ValueError(
+                f"Provided Laplacian {self.Laplacian} is a scalar Laplacian. "
+                f"The .apply_to_vector method is only suitable for vector Laplacians."
```
rabernat, line 333:
```suggestion
                f"The ``.apply_to_vector`` method is only suitable for vector Laplacians."
```

tests/test_kernels.py
```diff
@@ -211,3 +221,115 @@ def test_tripolar_exchanges(grid_type_field_and_extra_kwargs):
...
+    data_v = np.random.rand(ny, nx)
+
+    return grid_type, data_u, data_v, extra_kwargs, geolat_u
+
+
+def test_conservation_under_solid_body_rotation(
+    vector_grid_type_field_and_extra_kwargs,
+):
+    """This test checks that vector Laplacians are invariant under solid body rotations:
+    a corollary of conserving angular momentum."""
```
rabernat, line 295: Btw, I love this test! 🎉

rabernat (approved):
Fantastic. Really excited to try this.

rabernat:
I manually resolved the merge conflict and it made pre-commit mad. But I'm just going to merge this so we can move forward.

PR #36, opened 2021-05-03
Update `s_max`

iangrooms:
The current configuration tries to make the filter a close approximation to the target over a range of wavenumbers that is conservative in the sense that our second-order Laplacians don't access the full range of wavenumbers over which the filter is accurate. To reduce cost (number of stages) and improve stability, I've reduced the range of wavenumbers to one that is appropriate to the second-order Laplacians. (It's actually still conservative because it assumes the smallest grid spacing is the same in both directions.) I've also updated the stage order to reflect this; it should improve numerical stability by having a more-accurate estimate of which stages are strongly damping (or amplifying).


iangrooms:
Strange that it fails pre-commit, since it passes locally.

rabernat:
> Strange that it fails pre-commit

This is probably because we do not have a specific version of black pinned:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/f3f644a3b291f8ee0b5348b47aa74f14e80af0a9/.pre-commit-config.yaml#L36-L40

So your local version (probably installed months ago) differs from the CI version (installed today at whatever the latest version is). We could fix this by pinning a version in `.pre-commit.yaml`.

iangrooms:
I updated and re-ran, which is how I got it to pass. In the future I can just update, but if anyone new starts contributing it might be good to pin a specific version so they don't have to deal with these niggles.

rabernat (approved):
This seems reasonable to me!

iangrooms:
There's no big rush on this, so I'm going to leave this PR open for a day or two before merging, to see if anyone else comments.

NoraLoose:
Nice! 

Would it make sense to introduce a kernel parameter like `order` or `accuracy` (default = 2) and compute the target filter based on this information? Just in case someone contributes a higher-order Laplacian in the future?

iangrooms:
I do think that would be good to have, but not worth implementing until someone actually contributes a higher-order Laplacian. Anything we do now might end up getting revised anyways if/when someone contributes a higher-order Laplacian.

iangrooms:
My most recent change is extremely minor: flip the order of the amplifying stages so that most-amplifying come last. It has a huge beneficial impact in my idealized tests, so it would be good to try it out with some real data. I'm going to merge this PR now.

PR #57, opened 2021-05-05
Remove the test modules from coverage report

jbusecke:
We should base our coverage only on the actual module, not the test files. 

This also excludes some other python files which should not be tested.


jbusecke:
Closing in favor of #72

Issue #38, opened 2021-05-05
Add Filter Theory


jbusecke:

iangrooms:
@jbusecke or @rabernat can you tell me how to build the documentation locally so I can see a rendered version of my updates to the theory.rst file?

jbusecke:
See #60.

PR #60, opened 2021-05-06
Added how to build docs locally

jbusecke:
In #38 @iangrooms asked for instructions on how to build docs locally. I have added instructions here, but ironically I was not able to build them locally. I think there is a problem with my conda env (not resolving to the latest `myst_nb` version). 

@iangrooms could you try to follow these instructions and see if it works for you? Might just be some quirk of my local machine.

### DIFF
--- a/docs/how_to_contribute.rst
+++ b/docs/how_to_contribute.rst
@@ -67,7 +67,7 @@ How to change the documentation
 
 In order to build the documentation locally you should build and activate the docs environment::
 
-   conda env create -f docs/environment.yml
+   mamba env create -f docs/environment.yml
 
    conda activate gcm-filters-docs
 
--- a/docs/how_to_contribute.rst
+++ b/docs/how_to_contribute.rst
@@ -61,3 +61,24 @@ until all checks pass.
 
 
 Once you got everything to pass, you can stage and commit your changes and push them to the remote github repository.
+
+How to change the documentation
+-------------------------------
+
+In order to build the documentation locally you should build and activate the docs environment::
+
+   conda env create -f docs/environment.yml
+
+   conda activate gcm-filters-docs
+
+Then navigate to the docs folder and build the docs locally with::
+
+   cd docs
+
+   make html
+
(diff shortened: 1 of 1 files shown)


docs/how_to_contribute.rst
```diff
@@ -61,3 +61,24 @@ until all checks pass.
 
 
 Once you got everything to pass, you can stage and commit your changes and push them to the remote github repository.
+
+How to change the documentation
+-------------------------------
+
+In order to build the documentation locally you should build and activate the docs environment::
+
+   conda env create -f docs/environment.yml
```
rabernat, line 70:
```suggestion
   mamba env create -f docs/environment.yml
```

We already recommend mamba elsewhere, why not here?

docs/how_to_contribute.rst
```diff
@@ -61,3 +61,24 @@ until all checks pass.
...
+
+Then navigate to the docs folder and build the docs locally with::
+
+   cd docs
+
+   make html
+
+Once that is done you can open the created html files in `docs/_build/index.html` with your webbrowser::
+
+   open _build/index.html
```
rabernat, line 82:
```suggestion
   cd _build/html; python -m http.server
```

Better to preview the docs through an actual web server. [Sphinx autobuild](https://pypi.org/project/sphinx-autobuild/) is another good option.

iangrooms: Thanks! It works for me, though I do get this warning on the "make" command:

`WARNING: html_static_path entry '_static' does not exist`

Not sure what the `python -m http.server` command does. For me it appears to do nothing. But I can just open index.html with my browser instead.

docs/how_to_contribute.rst
jbusecke: Sure thing.

docs/how_to_contribute.rst
jbusecke: I am also just learning about this. @rabernat is there some documentation on this that I could read up on? Happy to change this though.

Issue #58, opened 2021-05-05
Bizarre CI failures


jbusecke:
I have experienced some really strange behavior in the CI for my last PRs (#54 #56 #57). 

In each of these, I did not change any code from the package itself, yet I am getting failures for different python versions. 

In one of the cases, the same test is executed twice, one time it [passes](https://github.com/ocean-eddy-cpt/gcm-filters/pull/57/checks?check_run_id=2511414883), the other time it [fails](https://github.com/ocean-eddy-cpt/gcm-filters/pull/57/checks?check_run_id=2511419641) for **literally** the same setup.

My first intuition was that there might be a (temporary) issue with the dependency resolution, but I diff'ed both of the logs and the versions are *exactly* the same. 

I am really not sure how to proceed with this. Any help would be greatly appreciated. 

cc @rabernat

jbusecke:
Is this really just a transient error? It seems like #36 went through and the build isn't failing anymore.

I restarted #54 #57 and will see if the problem just goes away.

jbusecke:
It did not resolve itself, but I have an idea what is going wrong.

`test/test_kernel.py` contains [these lines](https://github.com/ocean-eddy-cpt/gcm-filters/blob/4630768f487eb2eb5bcccbfae766c786f45b8fdb/tests/test_kernels.py#L131-L153), which are failing seemingly randomly.

The test generates a random `random_xloc` within the interval `[0, nx]`, but then you add 2 to the index? I think this could explain that these tests just randomly fail (if `random_xloc` ends up being `>nx-1`). It seems this was introduced [here](https://github.com/ocean-eddy-cpt/gcm-filters/blame/4630768f487eb2eb5bcccbfae766c786f45b8fdb/tests/test_kernels.py#L153) by @NoraLoose. Could you confirm if this is a bug, or if I am missing something?

If I understand this correctly simply changing `random_xloc = np.random.randint(0, nx)` to `random_xloc = np.random.randint(0, nx-2)` should do the trick?

NoraLoose:
@jbusecke, you are totally right. This is a bug that I introduced and hadn't noticed before. Sorry that this has caused such seemingly random and bizarre behavior! I will open a PR that fixes the test. Thanks for catching this!

jbusecke:
No problem, whatsover. Just glad we got to the bottom of it!

PR #61, opened 2021-05-06
Fix bug in test for irregular Laplacians

NoraLoose:
This fixes the bug that @jbusecke uncovered in https://github.com/ocean-eddy-cpt/gcm-filters/issues/58#issuecomment-832989721. Thanks, Julius!

I shortened the range within which`random_xloc` gets generated from [0, nx) to [2, nx-2) because later on in the test, 2 gets added/substracted to the index. Similarly for `random_yloc`.


NoraLoose:
This is such a minor change -- which was essentially made by @jbusecke --, so I will go ahead and merge this PR myself. Hope this will fix the CI issues (https://github.com/ocean-eddy-cpt/gcm-filters/issues/58).

PR #62, opened 2021-05-06
Docs theory

iangrooms:
Rather than reproduce the details in the paper, I aimed for a more high-level approach. Comments welcome. Fixes ocean-eddy-cpt/gcm-filters#38.


NoraLoose:
Do we want to keep the "Filter Theory" section completely free of code snippets? We could introduce an additional method to the Filter Class which plots the target and approximated filter (similar to Figure 2 in the paper). I could imagine to have some snippets like
```
filter = gcm_filters.Filter(
    filter_scale=4,
    dx_min=1,
    filter_shape=gcm_filters.FilterShape.TAPER,
    grid_type=gcm_filters.GridType.REGULAR,
)
filter.plot_shape()
```
between paragraphs.

I think this would be useful, but we could also defer it to a different place in the docs. Thoughts?

iangrooms:
I like the idea, and I'm planning on adding a method to plot the shape.

iangrooms:
I added some words about `plot_shape` to the theory, but didn't see a way to introduce code snippets without having to go into a much lengthier presentation of the code. I did add `plot_shape` to tutorial.ipynb in #66, which will show how to use and interpret the plot.

rabernat:
Great work Ian! I propose we merge this as is and do some further iterations via more PRs. I personally think that interspersing some actual code snippets with the theory would enhance understanding, but we can try adding that later.

In order to make this mergeable you will have to remove the "draft" status.

PR #63, opened 2021-05-06
Communicate `n_steps`

NoraLoose:
##  Purpose

This is a tiny PR which helps communicate `n_steps` to the user in case they specify `n_steps = 0` (which will trigger the computation of the default `n_steps`). 

## Changes

I simply moved the computation of the default `n_steps` to the post-init function in the Filter Class, which has the following effect on the repr of the filter object:
* **new**: tells the user the computed default `n_steps`;
*  **old**: tells the user `n_steps = 0` (not very useful).

I also re-ran one of the tutorials which used `n_steps = 0`; it now produces the correct string representation of the filter object that is created with `n_steps = 0`.


iangrooms:
Looks good to me. Perhaps this is bad etiquette, but I'd like to request adding a new feature as part of this pull request: warnings whenever 
1. The user manually sets `n_steps` lower than the default
2. The user selects a filter scale that is so much bigger than the grid scale that it may lead to numerical instability.

My main branch has a draft [here](https://github.com/iangrooms/gcm-filters/blob/bedaccbbd6d023a2d80f53ea57dbb7d38f6eef37/gcm_filters/filter.py#L70-L119), with everything still in the `_compute_filter_spec` function. (Moving it to `__post_init__` is a good idea.) Instead of the "if n-steps is zero, set it to default" approach, I compute the default regardless of the value input by the user for `n_steps`. Then if they used `n_steps=0` I just set `n_steps` to the default, and if they set `n_steps` _below_ the default I issue a warning.

We could merge this PR and then add the warnings separately, or I could try to edit this PR (haven't done that before). Let me know.

NoraLoose:
Thanks for looking at this!

I think your suggested changes sound good too. Probably less messy to open a separate PR after this is merged, unless you want to check out this branch (`NoraLoose:communicate-nsteps`) and push to it. Both works - up to you.

iangrooms:
I'll just merge this and then build the warnings on top of it.

PR #64, opened 2021-05-07
Add warnings

iangrooms:
This PR adds two sets of warnings:

1. If `n_steps` is manually set below the default
2. If the filter factor is large enough that numerical instability might be a problem.

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -1,5 +1,6 @@
 """Main Filter class."""
 import enum
+import warnings
 
 from dataclasses import dataclass, field
 from itertools import chain, zip_longest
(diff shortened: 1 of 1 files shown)


NoraLoose:
Related to what @jbusecke said, here are specific examples from our code (which test the raising of errors rather than warnings):

https://github.com/ocean-eddy-cpt/gcm-filters/blob/282a36f7ec0d779723e8d1b698d2835d1dacb09f/tests/test_kernels.py#L191-L192

https://github.com/ocean-eddy-cpt/gcm-filters/blob/282a36f7ec0d779723e8d1b698d2835d1dacb09f/tests/test_filter.py#L144-L147

rabernat:
I am a big fan of these warnings. Thanks for adding this Ian.

I just looked at the code and realized that we are essentially encoding a lot of "data" about the filters via a bunch of if / else statements. This predates this PR, e.g.

https://github.com/ocean-eddy-cpt/gcm-filters/blob/282a36f7ec0d779723e8d1b698d2835d1dacb09f/gcm_filters/filter.py#L227-L235

but this PR adds more of the same.

It occurred to me that a much more compact way to do this would be to have a standalone data structure with all this information, e.g.

```python
filter_params = {
    FilterShape.GAUSSIAN:
        {1: {'size': 0.8, 'max_filter_factor': 67},
         2: {'size': 1.1, 'max_filter_factor': 77}
    # etc
}
```

That would keep all of the information in one place and separate it from the actual logic of the Filter class.

Thoughts?

iangrooms:
I added the standalone data structure per @rabernat's suggestion. 

Adding tests to check the warnings is a bit beyond my Python skills. The current tests already generate a warning because `n_steps` in the test is below the default. Maybe if you could show me how to "catch" that warning I could then add code to test the other warnings too.

Finally, for some reason pre-commit passes locally (except flake8, which always fails), but not on github. I have run `pre-commit autoupdate` locally, which doesn't seem to help.

NoraLoose:
@iangrooms, I opened a PR to your branch `iangrooms:add-warnings` which adds tests for the filter warnings (and a few other minor fixes). If you approve and merge, the new tests should show up in this PR too.

NoraLoose:
@iangrooms I will leave this PR for someone else to review since I have contributed to the code myself.

.pre-commit-config.yaml
```diff
@@ -14,27 +14,27 @@ repos:
...
-  hooks:
-    - id: flake8
-      exclude: docs/source/conf.py
-      args: [--max-line-length=105, --ignore=E203,E501,W503, --select=select=C,E,F,W,B,B950]
+#- repo: https://gitlab.com/pycqa/flake8
+#  rev: 3.9.1
+#  hooks:
+#    - id: flake8
+#      exclude: docs/source/conf.py
+#      args: [--max-line-length=105, --ignore=E203,E501,W503, --select=select=C,E,F,W,B,B950]
```
rabernat, line 22: Noting that flake8 has ben disabled. We will need to sort out the pre-commit at a later point so we can reduce contributor friction.

rabernat (approved):
This is great and very thorough. Really appreciate the efforts.

.pre-commit-config.yaml
iangrooms: Thanks to @NoraLoose for cleaning up my code and adding the warnings! I got pre-commit to pass, but not sure how. `isort` seemed to be unable to make up its mind about where things should be.

NoraLoose:
@rabernat the test failure should be fixed if you merge https://github.com/ocean-eddy-cpt/gcm-filters/pull/65. Sorry about that!

rabernat:
> If #65 gets merged first, then I'm guessing it will fix this one automatically.

I did just merge #65. But there is one more step: you need to manually merge the upstream/master branch to your feature branch. Usually this goes something like this:
```python
git fetch upstream
git merge upstream/master
git push origin add-warnings 
```

jbusecke:
> @jbusecke could you disable the pre-commit check requirement for now? I don't want folks getting bogged down with these very nit-picky, version-related pre-commit problems. Ian and Nora have done a lot of good work here, their pre-commit passes locally. LGTM.

Sure. It is curious to me why this happens only here (repeatedly), but happy to kill it.

PR #66, opened 2021-05-10
Add method to plot filter shape

iangrooms:
This PR adds a new method to the filter class: `plot_shape`. It plots the shape of the target filter and the approximate filter in "Fourier space", i.e. not the shape of the kernel since that depends on the grid. I added some cells to the tutorial notebook showing how to use it, and I plan to mention it in the "Filter Theory" part of the docs (#62). 

To plot the filter shape you need a lot of things that are already computed within `_compute_filter_spec`. It seemed a bad idea to copy/paste that code, so I instead modified `_compute_filter_spec` so that it takes a boolean argument `plot_shape`. When `plot_shape=False` it acts the same as previously. When `plot_shape=True` it doesn't compute or return the full filter spec; instead, it plots the shape and returns nothing. The filter class method `plot_shape` just calls `_compute_filter_spec` with the argument `plot_shape=True`. 

I figure we should add some tests, but I unfortunately have no idea how to test plots.


iangrooms:
As far as I can tell, the tests fail because matplotlib.pyplot is not part of the required packages. (Naturally it passes locally, because I have matplotib installed.) Should I add matplotlib to `requirements.txt`?

rabernat (commented):
I would like to suggest a different way of implementing this.

It would be great if we could write a standalone function that would just plot a `FilterSpec`. For example
```python
plot_filter_spec(my_filter_spec)
```
We could put this in a new module called `utils.py`. 

To do this we may need to augment the `FilterSpec` type to include a few extra pieces of data. However, I feel that clearly separating plotting utilities from the core computational part of the code will be easier to maintain in the long run.

gcm_filters/filter.py
```diff
@@ -6,6 +6,7 @@
 from itertools import chain, zip_longest
 from typing import Iterable, NamedTuple
 
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
rabernat: A standalone plotting function call call `import matplotlib.pyplot as plt` from within the function definition to avoid a package-wide dependency.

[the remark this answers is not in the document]
iangrooms: I though you might not want to include matplotlib. I'll have to work on re-doing this.

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
             )
         self.grid_ds = xr.Dataset({name: da for name, da in self.grid_vars.items()})
 
+    def plot_shape(self):
```
rabernat, line 288:
```suggestion
    def plot_shape(self, ax=None):
```

Let's add an option to pass an existing axis object, in case the user wants to put the plot into a broader figure.

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
...
+        """Plot the shape of the target filter and approximation."""
+        # Plot the target filter and the approximate filter
+        s_max = self.filter_spec.s_max
+        target_spec = TargetSpec(s_max, self.filter_scale, self.transition_width)
+        F = _target_function[self.filter_shape](target_spec)
+        x = np.linspace(-1, 1, 10001)
+        k = np.sqrt(s_max * (x + 1) / 2)
+        plt.plot(k, F(x), "g", label="target filter", linewidth=4)
```
rabernat, line 296:
```suggestion
        if ax is None:
            fig, ax = plt.subplots()
        ax.plot(k, F(x), "g", label="target filter", linewidth=4)
```

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
...
+        plt.plot(
```
rabernat, line 297:
```suggestion
        ax.plot(
```

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
...
+            k,
+            np.polynomial.chebyshev.chebval(x, self.filter_spec.p),
+            "m",
+            label="approximation",
+            linewidth=4,
+        )
+        plt.axvline(
```
rabernat, line 304:
```suggestion
        ax.axvline(
```

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
...
+            2 * np.pi / self.filter_scale,
+            color="k",
+            label="filter cutoff wavenumber",
+            linewidth=2,
+        )
+        plt.xlim(left=0)
```
rabernat, line 310:
```suggestion
        ax.set_xlim(left=0)
```

gcm_filters/filter.py
```diff
@@ -282,6 +285,38 @@ def __post_init__(self):
...
+        if self.filter_scale / self.dx_min > 10:
+            plt.xlim(right=4 * np.pi / self.filter_scale)
+        bottom, top = plt.ylim()
+        plt.ylim(bottom=-0.1)
+        plt.ylim(top=1.1)
```
rabernat, line 315:
```suggestion
        ax.set_ylim(top=1.1)
```

rabernat (commented):
Ok great, this looks much cleaner.

We do need at least  "smoke test" for this: just a test that calls the `plot_shape` function and verifies that it doesn't error.

rabernat:
Fantastic!

tests/test_filter.py
```diff
@@ -144,6 +171,7 @@ def grid_type_and_input_ds(request):
 def test_filter(grid_type_and_input_ds, filter_args):
     grid_type, da, grid_vars = grid_type_and_input_ds
     filter = Filter(grid_type=grid_type, grid_vars=grid_vars, **filter_args)
+    filter.plot_shape()
```
rabernat, line 174: To me, this is a good enough "smoke test" for now. We call the function and make sure it doesn't error.

PR #67, opened 2021-05-10
Tutorial on numerical instability and coarsening

hmkhatri:
This PR adds a short tutorial on how to deal with filtering for large values of scale_factor, in which numerical instabilities can cause errors. This tutorial uses a combination of filtering and coarsening to filter over a wide range of scales.

Comments and suggestions are welcome.

Also, can anyone guide me on uploading data (used in tutorials) to the cloud, so users can run the example scripts?


iangrooms:
Thanks for this tutorial @hmkhatri! A few comments:
1. Are you using the latest version of the code? I was hoping that we could stably achieve a filter factor of 40 with the Gaussian filter using the latest version. 
2. Would it be possible to add to the example by interpolating the filtered field back to the full resolution? This is needed for people who want to access the eddy field, not just the filtered field.

hmkhatri:
@iangrooms Thanks.

1. Yes, I am using the latest version of the code. I never managed to achieve good accuracy with a filter factor = 40. Does numerical stability depend on the spectrum of field that is being filtered? In the example, I filtered vorticity. If I instead filter velocity or SSH, scale factor = 40 might work then. 
2. Sure, I will add code to interpolate the coarsened field back on the original grid.

iangrooms:
> Does numerical stability depend on the spectrum of field that is being filtered?
Yes, since the problem is associated with amplification at small scales, fields that have more small-scale energy will be more susceptible to instability. I was still hoping vorticity would be stable out to a filter factor of 40, but evidently not. Thanks!

hmkhatri:
@iangrooms As per your suggestion, now I interpolate coarsened data back to the original grid in the tutorial.

iangrooms:
This looks good to me. Thanks @hmkhatri! 

We have an issue (#40) as part of the to-do for the JOSS paper to switch all tutorials to cloud data. @NoraLoose do you think @hmkhatri should try to make this tutorial use cloud data before accepting this pull request?

I also noticed that the data being filtered is single precision. Promoting it to double before filtering would help with numerical stability. I will add a comment to the "Filter Theory" section about this.

NoraLoose:
@hmkhatri, [here](https://github.com/ocean-eddy-cpt/cpt-data) is a guide written by @rabernat on how to upload your data to the cloud (OSN). I have only been using cloud data that is part of [this catalog ](https://github.com/ocean-eddy-cpt/cpt-data/blob/master/catalog.yaml) already, but haven't uploaded any data myself. So maybe someone else can help?

hmkhatri:
@iangrooms Thanks for testing the notebook. I didn't realize that changing the precision would improve filtering. I have put another section on precision. I guess it is clear now that users can either use double-precision or coarsening to improve accuracy.

PR #68, opened 2021-05-12
Documentation of filter types

### LINKED ISSUE #40: Switch Tutorials to cloud data

### LINKED ISSUE #42: Add tutorial/documentation explaining use of `kappa` (anisotropy & varying filter scale)

NoraLoose:
This adds a new tutorial which extends and updates `tutorial_irregular_grid.ipynb` by several aspects:

* switch to cloud data (fixes #40)
* add examples for anisotropic filtering and spatially-varying filters (fixes #42)
* remove examples that use GPU for filtering (will be included in separate tutorial because this tutorial is quite long already)

The tutorial is dependent on https://github.com/ocean-eddy-cpt/gcm-filters/pull/28 being merged since it uses the `kappa`'s for anisotropic and spatially-varying filtering.

I had to comment out the check for large files in the pre-commit config to be able to submit this new tutorial. Is there a better solution?


iangrooms (commented):
Looks good to me. I found one typo " the grid smalling is the smallest". The examples using kappa make me realize that all the stage-order/stability arguments go out the window when the filter factor is varying all over the domain. But I think it would be pretty hard to pick a single stage order that would somehow strike a balance for good stability over the whole domain, so I don't think we need to change anything.

NoraLoose:
Thanks for looking through the tutorial, @iangrooms! I fixed the typo. 

> But I think it would be pretty hard to pick a single stage order that would somehow strike a balance for good stability over the whole domain, so I don't think we need to change anything.

Agreed. :+1:

rabernat:
More great work. Thanks for this Nora!

(For some reason, reviewNB is not able to post my comments, so posting them here directly.)

### Data

Is the catalog.yaml file public? I think not. So how will a user be able to run this notebook?

I propose we make the cpt-data repo public. And point directly at the catalog url here: https://raw.githubusercontent.com/ocean-eddy-cpt/cpt-data/master/catalog.yaml.

### "Simple Fixed Factor Filter"

We might want to consider implementing this internally within gcm-filters rather than requiring the user to jump through these hoops manually..

NoraLoose:
Thanks @rabernat for your comments! 

> Is the catalog.yaml file public? 

I copied it over from the cpt-data repo to the gcm-filters repo (into the docs folder), but making the cpt-data repo public would probably be cleaner.

> We might want to consider implementing this internally within gcm-filters rather than requiring the user to jump through these hoops manually..

If we want to implement this internally, this would have to happen in the `filter.py` module (rather than the `kernels.py` module) because 
* multiplication by `area` happens before applying the Laplacian for the first time; and
* division by `area` happens after applying the Laplacian for the last time.

We would want to do this only for a subset of Laplacians: `REGULAR`, `REGULAR_WITH_LAND`, `TRIPOLAR_REGULAR_WITH_LAND`, and all of these would need the additional required grid variable `area`. 

Are there any disadvantages of implementing this, e.g., can we imagine situations where the user wants to use the `REGULAR` Laplacian without the area-weighting? I can't think of any, but maybe someone else can?

iangrooms:
I think if the grid is really regular, then there's not much need to use gcm-filters, so it seems like a good idea to me to make `REGULAR` into simple-fixed-factor.

NoraLoose:
The only remaining issue with this tutorial is that simple fixed factor filtering should be done internally by `gcm-filters`. I opened issue #71, but may not get to implementing/solving it right now. Do we want to merge this tutorial in the meantime, and update it once #71 is solved?

iangrooms:
> Do we want to merge this tutorial in the meantime, and update it once #71 is solved?

That makes sense to me. Use different PRs for different problems: Add tutorial with this PR; updated REGULAR/simple fixed factor with a separate PR.

Issue #71, opened 2021-05-20
Handle simple fixed factor filtering internally


NoraLoose:
**Simple fixed factor filtering** requires the following three steps by the user:

1. Before filtering, multiply the field by the local cell `area`.
2. Apply `gcm-filters` filter with `dx_min=1 `and `filter_scale= desired fixed factor`, pretending the grid was uniform.
3. After filtering, divide filtered field by local cell `area`.

Only the second step is handled by `gcm-filters` (see [this tutorial](https://gcm-filters.readthedocs.io/en/latest/tutorial_irregular_grid.html) or the new tutorial in #68).

To avoid confusion among users, we could let `gcm-filters` also handle steps 1 and 3. Implementing steps 1 and 3 internally would have to happen in the `filter.py` module (rather than the `kernels.py` module) because
* multiplication by `area` happens before applying the Laplacian for the first time; and
* division by `area` happens after applying the Laplacian for the last time.

We would want to do this only for a subset of Laplacians: `REGULAR`, `REGULAR_WITH_LAND`, `TRIPOLAR_REGULAR_WITH_LAND`, and all of these would need the additional required grid variable `area`.

PR #75, opened 2021-06-11
Handle simple fixed factor filtering internally

NoraLoose:
### Issue
This PR fixes issue #71. 

For applying a simple fixed factor filter (as for example described in the "Anisoptropic Filtering" section of the [Filter Theory](https://gcm-filters.readthedocs.io/en/latest/theory.html)), the user was required to manually go through the following steps:
1. Before filtering, multiply the field by the local cell area.
2. Apply gcm-filters filter with dx_min=1 and filter_scale= desired fixed factor, pretending the grid was uniform.
3. After filtering, divide filtered field by local cell area.

The first step is essentially a coordinate transformation where the original (locally orthogonal) grid is transformed to a uniform Cartesian grid with dx=dy=1. The third step is the reverse coordinate transformation.

### Changes
These steps are now handled internally by the code.

I introduced a new Laplacian base class `BaseScalarLaplacianWithArea` for the Laplacians that are for simple fixed factor filtering: 
* `TRANFORMED_TO_REGULAR`
* `TRANSFORMED_TO_REGULAR_WITH_LAND`
* `TRIPOLAR_TRANSFORMED_TO_REGULAR_WITH_LAND`. 

Steps 1 and 3 from above are handled as part of the filter class for all Laplacians that are a subclass  of `BaseScalarLaplacianWithArea`.

### Tutorial changes
I updated all tutorials to reflect the new way of doing simple fixed factor filtering.

**Old**:
```
filtered = filter.apply(field * area, dims=['yh', 'xh']) / area
```
**New**:
```
filtered = filter.apply(field, dims=['yh', 'xh']) 
```

While updating the tutorials, I also fixed some typos and clarified some statements. Thanks @sdbachman for reading through the documentation and for providing these comments.


iangrooms (approved):
These changes look good to me. @rabernat might want to take a look for Python style issues.

gcm_filters/filter.py
```diff
@@ -280,6 +286,19 @@ class Filter:
...
+        # Determine whether this is simple fixed factor filter; in that case we need dx_min = 1
+        if issubclass(self.Laplacian, BaseScalarLaplacianWithArea):
+            if self.dx_min != 1:
+                warnings.warn(
+                    "Provided Laplacian is for simple fixed factor filtering, "
+                    "where area-weighted field is filtered on a regular grid with dx = dy = 1 "
+                    "--> dx_min is set to 1",
+                    stacklevel=2,
+                )
+                self.dx_min = 1
```
rabernat, line 300: What scenario are you imagining here with this check / warming? Rather than overwriting `dx_min`, why don't we just raise a `ValueError` here and make the user fix it explicitly?

gcm_filters/filter.py
```diff
@@ -309,8 +328,8 @@ def __post_init__(self):
         ]
         if filter_factor >= max_filter_factor:
             warnings.warn(
-                "Warning: Filter scale much larger than grid scale -> numerical instability possible",
-                UserWarning,
+                "Filter scale much larger than grid scale -> numerical instability possible",
```
rabernat, line 331: Can we link to a documentation page with more context?

gcm_filters/filter.py
```diff
@@ -369,11 +387,14 @@ def plot_shape(self, ax=None):
...
         """Filter a field with scalar Laplacian across the dimensions specified by dims."""
-        if not issubclass(self.Laplacian, BaseScalarLaplacian):
+        if issubclass(self.Laplacian, BaseVectorLaplacian):
             raise ValueError(
                 f"Provided Laplacian {self.Laplacian} is a vector Laplacian. "
                 f"The ``.apply`` method is only suitable for scalar Laplacians."
             )
+        if issubclass(self.Laplacian, BaseScalarLaplacianWithArea):
+            # simple fixed factor filtering multiplies field by area before filtering
+            field = field * self.grid_ds["area"]
```
rabernat, line 397: Rather than putting this login in the `filter` module, why not put it into the kernel itself?

gcm_filters/kernels.py
```diff
@@ -38,6 +39,36 @@ def _prepare_tripolar_exchanges(field):
...
+    by a multiplication / division by the cell area of the original grid. Note that this occurs at the
+    filter level, rather than the Laplacian level.
+
+    Attributes
+    ----------
+    area: cell area
+    """
+
+    area: ArrayType
+
```
rabernat, line 71: There should be a way to make the class itself do the area weighting / deweighting. To me that would be a lot cleaner than doing it manually in the `filter` module (better separation of concerns).

rabernat (commented):
First, let me apologize deeply and sincerely for taking so long to review this. The past month has been extremely challenging for me work-wise due to family and travel-related reasons.

I really like the spirit of this PR. I think it makes things much simpler to implement the weighting internally. However, I have one large-ish suggestions for the implementation.

I don't think it makes sense to check the type of the kernel in `filters.py` and then do things differently based on what we find. That is a coding pattern that indicates poor separation of concerns between modules. Instead, I think kernels.py should take care of the weighting.

I think the best way to accomplish this would be to expand the Kernel interface to include `prepare` and `finalize` methods in the BaseLaplacian classes.

The default would be to do nothing, e.g.:

```python
    def prepare(self, field):
        return field
```

But implementations could override this, e.g.
```python
@dataclass
class AreaWeightedLaplacian(BaseLaplacian)
    area: ArrayLike

    def prepare(self, field):
        return field * self.area

    def __call__(self, field):
        # do stuff

    def finalize(self, field):
        return field / self.area
```

This could even be a mixin, so we don't have to redefine the core laplacian `__call__` functions, e.g.

```python
@dataclass
class AreaWeightedMixin(BaseLaplacian)
    area: ArrayLike

    def prepare(self, field):
        return field * self.area

    def finalize(self, field):
        return field / self.area

@dataclass
class RegularLaplacianWithArea(AreaWeightedMixin, RegularLaplacian):
    pass
```

Then `filter.py` would just call `field = kernel.prepare(field)`, rather than implementing the weighting directly.

Does this suggestion make sense? Happy to chat more, and sorry again for my slowness.

I really appreciate your diligent and careful work on this project.

NoraLoose:
> Instead, I think kernels.py should take care of the weighting.

> I think the best way to accomplish this would be to expand the Kernel interface to include prepare and finalize methods in the BaseLaplacian classes.

Thanks for this suggestion, Ryan! I really like it; it makes things way cleaner. 

I updated the PR, and incorporated all your comments and suggestions.

gcm_filters/kernels.py
```diff
@@ -84,9 +122,31 @@ def __call__(self, field: ArrayType):
...
+    ----------
+    area: cell area of original grid
+    """
+
+    area: ArrayType
+
+    pass
+
+
+ALL_KERNELS[GridType.TRANSFORMED_TO_REGULAR] = RegularLaplacianWithArea
```
rabernat, line 144: I find this naming confusing. "Transformed" is ambiguous, and it doesn't indicate anything about the area weighting. What about`REGULAR_AREA_WEIGHTED`?

This is not a dealbreaker for me...but I'm just trying to think about what is most obvious for users.

gcm_filters/kernels.py
```diff
@@ -126,9 +185,37 @@ def __call__(self, field: ArrayType):
...
+    """
+
+    area: ArrayType
+    wet_mask: ArrayType
+
+    pass
+
+
+ALL_KERNELS[
+    GridType.TRANSFORMED_TO_REGULAR_WITH_LAND
```
rabernat, line 212: Same. Do we really want to call this "TRANSFORMED"? Or would "AREA_WEIGHTED" be more clear?

gcm_filters/kernels.py
```diff
@@ -231,29 +326,33 @@ def __call__(self, field: ArrayType):
...
 
         out = out[..., :-1, :]  # disregard appended row
 
         out = self.wet_mask * out
         return out
 
 
-ALL_KERNELS[GridType.TRIPOLAR_REGULAR_WITH_LAND] = TripolarRegularLaplacianTpoint
+ALL_KERNELS[
+    GridType.TRIPOLAR_TRANSFORMED_TO_REGULAR_WITH_LAND
```
rabernat, line 338: Same.

rabernat (commented):
Thanks for the super quick revisions! 🚀 

One more round of comments, focused on the naming of stuff. Then LGTM!

gcm_filters/kernels.py
NoraLoose: I actually went back and forth with the naming convention here; something similar to what you suggest here has been in the mix too! So I'm glad to hear that you find `AREA_WEIGHTED` most intuitive.

gcm_filters/kernels.py
```diff
@@ -193,14 +280,22 @@ def __call__(self, field: ArrayType):
...
+    boundary. There are three steps:
+    1) Field on locally orthogonal grid is transformed to field on regularly spaced Cartesian
+    grid with dx = dy = 1, through multiplication by cell area of original grid.
+    2) Laplacian acts on regular Cartesian grid.
+    3) Diffused field on regular Cartesian grid is transformed back to field on original grid,
+    through division by cell area of original grid.
 
     Attributes
     ----------
+    area: cell area of original grid
```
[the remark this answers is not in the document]
NoraLoose: True, except that the function ` required_grid_vars` will not pick up `area` if I don't redefine it here. 

This is only an issue for the classes `RegularLaplacianWithLandMaskAndArea` and `TripolarRegularLaplacianTpoint` where additional attributes have to be defined (in addition to what is inherited from the superclasses). In contrast, it is not an issue for the `RegularLaplacianWithArea` which does not need any additional attributes.

gcm_filters/kernels.py
```diff
@@ -84,9 +122,31 @@ def __call__(self, field: ArrayType):
...
 
 
+@dataclass
+class RegularLaplacianWithArea(AreaWeightedMixin, RegularLaplacian):
+    """̵Scalar Laplacian operating on a locally orthogonal grid in three steps:
+    1) Field on locally orthogonal grid is transformed to field on regularly spaced Cartesian
+    grid with dx = dy = 1, through multiplication by cell area of original grid.
+    2) Laplacian acts on regular Cartesian grid.
+    3) Diffused field on regular Cartesian grid is transformed back to field on original grid,
+    through division by cell area of original grid.
```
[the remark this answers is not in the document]
NoraLoose: Thanks - the lists render correctly now. 

I also tried to get rid off the hyphen that appears in the API at the beginning of the rendered docstrings for some of the Laplacians (but not for others). I couldn't resolve this issue. Any ideas?

NoraLoose:
Since I changed the names of 3 Laplacians (from `TRANSFORMED_TO` to `AREA_WEIGHTED`) I should also update all the example notebooks for the documentation. I will do that as soon as casper is up and running again.

rabernat:
> I could also do the notebook update as part of PR #78,

Let's go with that! I think we should merge this now.

gcm_filters/kernels.py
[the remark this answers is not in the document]
rabernat: I do not, but this is relatively minor, so not a big deal to me.

gcm_filters/kernels.py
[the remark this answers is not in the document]
rabernat: Ok that makes sense. I don't really understand why `return list(self.__annotations__)` in `BaseScalarLaplacian` does not pick up on the mixin attributes; it must have to do with the nitty-gritty details of class inheritance. Perhaps fixable somehow in the `required_grid_args` but not worth more effort here.

Issue #77, opened 2021-06-17
What is still missing in the documentation?


NoraLoose:
- [ ] #37 Statement of need (on landing page, above table of contents; can be similar to what we write in the JOSS paper: #76)
- [ ] #39 Installation instructions
- [ ] What else?

Right now, we have a page with the **Filter Theory** and **many tutorials**. 
* Do people feel that we need something **in between**, like a general introduction to how the package works, e.g., a mention of the core object `gcm_filters.Filter`? 
* A lot of the important general concepts are covered in the [Filter Theory](https://gcm-filters.readthedocs.io/en/latest/theory.html) and the [first tutorial](https://gcm-filters.readthedocs.io/en/latest/tutorial.html). So a second option is to leave the documentation as is (with potential small edits of the Filter Theory and the first tutorial), since folks prefer to learn with examples anyway.

Thoughts? The more opinions, the better.

rabernat:
Thanks so much for continuing to work on this Nora! It's really coming together. I agree we should make some tweaks to the documentation organization.

Regarding installation, I have just released v0.1 pypi, which means you can just `pip install gcm-faces`. We should feel free to keep making new releases as often as we want when new features com online.

> * Do people feel that we need something **in between**,

YES. I think we would benefit what is called "narrative documentation", which explains how to use the code in more verbose detail. This would discuss the syntax, the Filter object, the options, etc. It might be possible to adapt the first tutorial to serve this role. Ideally some of the redundancy between the tutorials could be eliminated by subsuming some other concepts into this page (e.g could we explain GPU filtering on that page). Many of the tutorials are halfway between narrative documentation for different features and examples.

A "quickstart" page might also be useful for people who just want to copy and paste something and get started immediately.

I also think we should rename the other "tutorials" as "examples" and put them in a sub directory. The new layout would be like this

```
docs/examples/index.rst  # links out to all of the examples
docs/examples/example_filter_types.ipynb
docs/examples/...
```

Above all, whatever organization we choose, the top level page (https://gcm-filters.readthedocs.io), should describe the structure and organization of the documentation, so users can find what they are looking for more easily.

PR #78, opened 2021-07-09
Docs update

NoraLoose:
This is a first attempt to restructure the documentation, with the goal of making it more user-friendly. As discussed / suggested in #77, I have made the following changes:

* Add short summary of GCM-Filters on landing page (`index.rst`)
* Add installation instructions
* Convert the first tutorial to a narrative documentation (`basic_filtering.rst`)
* As part of `basic-filtering.rst`, discuss the filter object and all grid types
* Add code snippets to `theory.rst` (essentially the `.plot_shape` plots)
* Rename tutorials --> examples, and move them to their own directory

Still to do: 
* A quickstart page?
* ...


rabernat:
Let me know how I can best be helpful here. Is this ready for a full read-through?

NoraLoose:
The name change `TRANSFORMED_TO` --> `AREA_WEIGHTED` still has to be updated in all example notebooks. I'm still waiting for casper to go online again.

But otherwise, this is ready for a read-through. The big changes are in the `.rst` files (rather than the notebooks) anyway.

docs/theory.rst
```diff
@@ -16,79 +22,188 @@ The filter shape can be thought of in terms of the kernel of a convolution filte
 
 where :math:`f` is the function being filtered, :math:`G` is the filter kernel, and :math:`x'` is a dummy integration variable.
```
iangrooms, line 23: Per Hussein's comments on our paper, I think we could add a parenthetical statement here saying that our filter with a Gaussian target does not produce _exactly_ the same results as convolution against a Gaussian kernel on the sphere.

docs/theory.rst
```diff
@@ -16,79 +22,188 @@ The filter shape can be thought of in terms of the kernel of a convolution filte
...
+
+    gaussian_filter = gcm_filters.Filter(
+        filter_scale=4,
+        dx_min=1,
+        filter_shape=gcm_filters.FilterShape.GAUSSIAN,
+        grid_type=gcm_filters.GridType.REGULAR,
+    )
+    gaussian_filter
+
+Once the filter has been constructed, the method ``plot_shape`` can be used to plot the shape of the target filter and the approximate filter.
```
iangrooms, line 47: Suggestion: Add a comment saying that the distinction between the target filter and the approximate filter will be discussed below.

docs/theory.rst
```diff
@@ -16,79 +22,188 @@ The filter shape can be thought of in terms of the kernel of a convolution filte
...
 If the filter scale is much larger than the grid scale, many steps are required.
-Also, the Taper filter requires more steps than the Gaussian filter for the same ``filter_scale``.
+Also, the Taper filter requires more steps than the Gaussian filter for the same ``filter_scale``; in the above examples the Taper filters required ``n_steps`` = 16, but the Gaussian filter only ``n_steps`` = 5.
 
-The code allows users to set their own number of steps ``n_steps``.
-The minimum number of steps is 3; if ``n_steps`` is not set by the user, or if it is set to a value less than 3, the code automatically changes ``n_steps`` to the default value.
-If the number of steps is set too low the filter will not behave as expected: it may not have the right shape or the right length scale.
-
-Biharmonic steps are counted as 2 steps because their cost is approximately twice as much as a laplacian step.
+The code allows users to set their own ``n_steps``.
```
iangrooms, line 128: Somewhere in here we should say that the default number of steps computed by the code for the Taper filter assumes the default transition width. If the user selects a wider transition, then they might want to set a value of `n_steps` lower than the default to save cost while keeping accuracy. If the user selects a narrower transition then they should set a value of `n_steps` higher than the default, otherwise the filter won't behave as expected.

iangrooms (commented):
Overall this looks great! A big improvement that I'm sure took a lot of work. I just left a couple of comments on the filter theory section, but otherwise it looks good to me.

docs/theory.rst
NoraLoose: Would it be straightforward to change the code to the effect that it takes the input transition width into account when computing the default `n_steps`?

rabernat:
Will do my very best to review this early next week. Thanks for all your work.

docs/theory.rst
iangrooms: I set the scaling empirically, so I guess I could look for some kind of empirical scaling law that takes the transition width into account. I will put it on my to do list for the next few days.

rabernat (approved):
Finally got around to reviewing this. It's a very impressive update to the documentation, a huge improvement of where we were before. I have not more suggestions beyond what has already been done. Thanks so much @NoraLoose for your work!

I resolved a merge conflict in the `Filter` class docstring. Once CI passes, I will merge this.

Issue #45, opened 2021-05-05
Performance evaluation


jbusecke:
JOSS askes for `If there are any performance claims of the software, have they been confirmed? (If there are no claims, please check off this item.)`

What should/can we do to document performance of the package?

NoraLoose:
We started the discussion on performance evaluation [here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/34#issuecomment-801235449) and [here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/4).

jbusecke:
Apologies for missing those. Should I close this issue?

NoraLoose:
Not at all! I just posted these, so we don't need to start the discussion from scratch. :)

jbusecke:
Perfect!

NoraLoose:
Sorry, I didn't mean to kill this discussion. I will try to summarize the aspects that we had discussed to include in a performance evaluation. 

* Compare performance of `gcm-filters` on **CPUs** and **GPUs** against fortran implementations of spatial filters, e.g., the ones implemented by @iangrooms and @arthurBarthe. [Here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/4#issuecomment-780801306) is a list compiled by @rabernat how such a comparison could be done.
* Look at [strong and weak scaling](https://www.kth.se/blogs/pdc/2018/11/scalability-strong-and-weak-scaling/), see @rabernat's comment [here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/34#issuecomment-801235449).
* Compare computational speed when using increasingly complex Laplacians in `gcm-filters`, see for example the table [here](https://github.com/ocean-eddy-cpt/gcm-filters-paper/issues/34#issuecomment-801208072).
* ...

I am sure this list is incomplete, so please feel free to add/delete items. Maybe @iangrooms also saved the computational section that was originally part of the JAMES paper.

**Question**: Do we want to include the performance evaluation

1. in the JOSS paper,
2. in the `gcm-filters` documentation,
3. both (in some kind of combination)?

I think 3. would maybe work best.

rabernat:
Thanks for the writeup. I think your list is perfect, and I favor option 3.

[...]

PR #79, opened 2021-07-27
Refactor kernel tests

rabernat:
Here I am starting to refactor the kernel tests in preparation for a possible refactoring of the kernels to use numba (see #45 for context).

Goals:
- [x] Make the test suite deterministic (exact same tests are always run). This is accomplished by using an explicit constant seed for the random number generator
- [x] Eliminate boilerplate and repetition via better test parameterization (still more to do here)
- [ ] Add actual verification tests, i.e. a given function gives a specific answer. This will be important to give us confidence in actually replacing the kernel functions.

I am not quite sure how to proceed with the last item. 

### Option 1: Verify summary statistics for laplacians

One option would be to save some sort of summary statistic for the computed laplacian fields, e.g.
```python
res = laplacian(data)
res_squared_sum = (res**2).sum()
np.testing.assert_allclose(res_squared_sum, expected)
```
where `expected` is a number we actually save explicitly into the test suite based on the current code. Here we are not checking every value but just some summary. The pro is that this is simple. The con is that multiple fields could give the same answer.

### Option 2: Verify the entire field.

```python
res = laplacian(data)
np.testing.assert_allclose(res, expected)
```

Here we are checking the entire 128 x 256 test field. The challenge here is that we would need to actually save an array somewhere for each kernel. The easiest place would be to commit directly to the repo. But putting binary data into a github repo sucks. We would need to choose a data format, generate the expected data, etc.

Thoughts?

### DIFF
--- a/tests/test_kernels.py
+++ b/tests/test_kernels.py
@@ -34,6 +34,7 @@ _extra_kwargs = {
     GridType.TRIPOLAR_POP_WITH_LAND: ["wet_mask", "dxe", "dye", "dxn", "dyn", "tarea"],
 }
 
+
 ################## Scalar Laplacian tests ##############################################
 @pytest.fixture(scope="module", params=scalar_grids)
 def grid_type_field_and_extra_kwargs(request):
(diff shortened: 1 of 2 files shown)


tests/test_kernels.py
```diff
@@ -76,105 +72,75 @@ def test_required_grid_vars(grid_type_field_and_extra_kwargs):
...
+            ),
+        },
+    }
+
+    var_to_modify, slice_left, slice_right = replace_data[grid_type][direction]
+    print(grid_type, var_to_modify, slice_left, slice_right)
+    new_data = np.ones_like(test_kwargs[var_to_modify])
+    new_data[slice_left] = 1000
+    new_data[slice_right] = 2000
+    test_kwargs[var_to_modify] = new_data
```
rabernat, line 130: @NoraLoose - I did a pretty big refactoring of these "flux" tests. I'd like to make sure you approve these changes and understand how the new code works.

tests/test_kernels.py
```diff
@@ -76,105 +72,75 @@ def test_required_grid_vars(grid_type_field_and_extra_kwargs):
...
+    if grid_type not in irregular_grids:
+        pytest.skip("This test is only for irregular grids")
+
+    # deploy mass at random location away from Antarctica: delta_{j,i}
+
+    delta = np.zeros_like(data)
+    ny, nx = delta.shape
+    # pick a location outside of the mask
+    random_yloc = 99
+    random_xloc = 225
```
rabernat, line 93: I didn't see what we gained by making this point random.

tests/test_kernels.py
NoraLoose: @rabernat, this looks great to me. Thanks for cleaning up the code!

tests/test_kernels.py
```diff
@@ -76,105 +72,75 @@ def test_required_grid_vars(grid_type_field_and_extra_kwargs):
...
 # would be: all grids that have len(required_grid_vars)>1 (more than just a wet_mask)
-irregular_grids = [
-    gt for gt in GridType if "IRREGULAR" in gt.name or "TRIPOLAR_POP" in gt.name
-]
+irregular_grids = [GridType.IRREGULAR_WITH_LAND, GridType.TRIPOLAR_POP_WITH_LAND]
 
 
-def test_flux_in_y_direction(grid_type_field_and_extra_kwargs):
+@pytest.mark.parametrize("direction", ["X", "Y"])
+def test_flux_in_direction(grid_type_field_and_extra_kwargs, direction):
```
NoraLoose, line 79: Minor detail: Now that the flux tests for the x- and y-directions are combined, we should change the docstring of this test function and remove anything specific to the y-direction test. We could also remove the `_in_direction` from the name of the test function.

tests/test_kernels.py
NoraLoose: Yes, agreed.

NoraLoose (commented):
Thanks for starting to refactor the kernel tests! I like how the code is much more compact now. 

I only have one comment: All Laplacians get tested on a uniform grid now, because you specify `area` equal to a field of ones. For the `test_conservation` test, it would however be beneficial to test the Laplacians for irregular `area` data as before (i.e., those Laplacians that need an area as grid variable.)

For example, while I was doing the refactoring in PR #75 this morning, the `test_conservation` test caught some bugs related to multiple inheritance. This test will detect whether a fixed factor filter area-weights a field before filtering (and area-deweights it after filtering) or if this step is missed due to calling the prepare / finalize methods of the wrong superclass.

NoraLoose:
> For example, while I was doing the refactoring in PR #75 this morning, the test_conservation test caught some bugs related to multiple inheritance. 

My bad - for this specific example the relevant *conservation test* was the one within `test_diffusion_filter` in the `test_filter.py` module. 

More generally speaking, as long as we test for irregular `area` data in the `test_filter.py` module, I am okay with simplifying the conservation tests in `test_kernels.py` as you suggested. If the Laplacian is non-conservative, the filter will most likely not be conservative either.

NoraLoose:
> More generally speaking, as long as we test for irregular area data in the test_filter.py module, I am okay with simplifying the conservation tests in test_kernels.py as you suggested. 

On second thought, I disagree with what I just said. I do think we should test with irregular grid data (including `area`, `dx`, `dy`) in `test_kernels.py`, in particular in the verification tests that make sure that the numba Laplacians do the same thing as the current ones (your item 3).

If we tested all Laplacians on regular grid data (`dx` = `dy` = 1), most of our kernels (all except for the tripolar and vector Laplacians) would give the *exact same answer* and would produce the same expected field / statistics. So no matter if we choose your option 1 or 2, the verification test would be a rather weak test. That's why I vote for creating some irregular grid data in the fixture ` grid_type_field_and_extra_kwargs` that can be used for both the conservation and the verification tests.

rabernat:
> hat's why I vote for creating some irregular grid data in the fixture ` grid_type_field_and_extra_kwargs` that can be used for both the conservation and the verification tests.

This looks like what you did here, as part of #75:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/05b58b9e2df7d770a4eda7dbcd21f0611b45f667/tests/test_kernels.py#L45-L50

So as long as I merge those changes into this PR, we should be good, right?

NoraLoose:
> So as long as I merge those changes into this PR, we should be good, right?

Right! I got confused, and thought the irregular / random grid data was already part of master, and that this PR would revert to uniform grids. So yes, all good.

tests/test_kernels.py
```diff
@@ -17,57 +21,67 @@
...
-        mask_data[: (ny // 2), : (nx // 2)] = 0
-        mask_data[0, :] = 0  #  Antarctica
-        extra_kwargs["wet_mask"] = mask_data
+    for name in _grid_kwargs[grid_type]:
+        if name == "wet_mask":
+            extra_kwargs[name] = _make_mask_data(shape)
+        elif "kappa" in name:
+            extra_kwargs[name] = np.ones(shape)
+        else:
+            extra_kwargs[name] = _make_irregular_grid_data(shape)
```
rabernat, line 79: Note here that all of the irregular grid variables (e.g. dx, dy, etc.) are now getting distinct random data.

tests/test_kernels.py
```diff
@@ -17,57 +21,67 @@
...
-        mask_data[0, :] = 0  #  Antarctica
-        extra_kwargs["wet_mask"] = mask_data
-        grid_data = 0.5 + np.random.rand(ny, nx)
-        extra_kwargs["dxe"] = grid_data
-        extra_kwargs["dye"] = grid_data
-        extra_kwargs["dxn"] = grid_data
-        extra_kwargs["dyn"] = grid_data
-        extra_kwargs["tarea"] = grid_data * grid_data
+        # is this important? If not, remove special case
+        extra_kwargs["tarea"] = extra_kwargs["dxe"] * extra_kwargs["dye"]
```
rabernat, line 84: Does the POP kernel (or other kernels for that matter) care about consistency between the grid variables?

tests/test_kernels.py
```diff
@@ -84,13 +98,17 @@ def test_conservation(grid_type_field_and_extra_kwargs):
...
+        area = 1
+    else:
+        area = extra_kwargs.get("area", None)
+        if area is None:
+            area = extra_kwargs.get("tarea", 1)
+
+    res = laplacian(data)
+
+    # currently failing only for TRIPOLAR_POP_WITH_LAND. Why?
+    np.testing.assert_allclose((area * res).sum(), 0.0, atol=1e-12)
```
rabernat, line 111: Any idea why this test would fail only for TRIPOLAR_POP_WITH_LAND?

tests/test_kernels.py
```diff
@@ -104,111 +122,42 @@ def test_required_grid_vars(grid_type_field_and_extra_kwargs):
...
-        if grid_type == GridType.TRIPOLAR_POP_WITH_LAND:
-            test_kwargs["wet_mask"] = ones.copy()
-            test_kwargs["wet_mask"][0, :] = 0  # Antarctica
-            test_kwargs["tarea"] = ones
-            test_kwargs["dye"] = ones
-            test_kwargs["dyn"] = ones
-            test_kwargs["dxe"] = ones
-            test_kwargs["dxn"] = ones.copy()
-            test_kwargs["dxn"][random_yloc - 2, :] = 1000
-            test_kwargs["dxn"][random_yloc + 1, :] = 2000
```
rabernat, line 150: Since the irregular grids are all now automatically being populated with random data, my thinking was that it should be unnecessary to do this sort of manual manipulation of the grid variables to create anisotropy.

tests/test_kernels.py
```diff
@@ -104,111 +122,42 @@ def test_required_grid_vars(grid_type_field_and_extra_kwargs):
...
-            test_kwargs["dyn"] = ones
-            test_kwargs["dxe"] = ones
-            test_kwargs["dxn"] = ones
-            test_kwargs["dye"] = ones.copy()
-            test_kwargs["dye"][:, random_xloc - 2] = 1000
-            test_kwargs["dye"][:, random_xloc + 1] = 2000
+    # check that delta function gets diffused isotropically in y-direction
+    np.testing.assert_allclose(
+        diffused[random_yloc - 1, random_xloc],
+        diffused[random_yloc + 1, random_xloc],
```
rabernat, line 151: I was thinking this should be true in general for any irregular grid. However, the tests are currently failing, e.g.

```
test_flux[GridType.IRREGULAR_WITH_LAND]
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-12
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 1.55178064
E       Max relative difference: 0.62167217
E        x: array(0.944359)
E        y: array(2.49614)
```

It doesn't make sense to me that the concentration could _ever_ be greater than 1, but that's what we see in the second value.

Any insights into what might be going wrong here?

rabernat (commented):
I merged #75 into this PR and continued my refactoring of the tests.

It's mostly working ok, but I have some questions about test failures. Would appreciate more guidance from @NoraLoose whenever you have a second.

tests/test_kernels.py
rabernat: Small update, in 51121b3 I modified this test to use a constant area=1 and also reduced the amplitude of the randomness. But the test is still failing, albeit with different values

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-12
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.18457325
E       Max relative difference: 0.20314357
E        x: array(1.093159)
E        y: array(0.908585)
```

I definitely feel stuck on this.

tests/test_kernels.py
NoraLoose: This is great. Previously, I had assigned identical random data to `dx` and `dy`, which made the conservation test for `TRIPOLAR_POP_WITH_LAND` accidentally pass (see *Side note for the curious* in https://github.com/ocean-eddy-cpt/gcm-filters/pull/80#issue-701806629).

tests/test_kernels.py
NoraLoose: No, none of the kernels cares about this. So I removed this in #80.

tests/test_kernels.py
NoraLoose: See https://github.com/ocean-eddy-cpt/gcm-filters/pull/80#issue-701806629

tests/test_kernels.py
NoraLoose: This test was designed for regular grid data, so I reverted back to an earlier version of yours in #80. For more details, see https://github.com/ocean-eddy-cpt/gcm-filters/pull/80#issue-701806629.

tests/test_kernels.py
NoraLoose: I think values can be greater than 1 if your are on a irregular grid. Here is an example of how a flux across a northern cell edge computed:
```
 nflux = ((np.roll(data, -1, axis=-2) - data) / self.dyn * self.dxn)
```
If `self.dyn < 1` and `self.dxn >1`, then `nflux` will be larger than 1 if you consider the northern edge of the grid cell that the delta function lives in. The grid cell to the north does not get any flux contributions from other neighboring cells, so the concentration in this cell will be greater than 1 (for `area = 1`). Definitely wouldn't happen on a regular grid.

rabernat:
I am almost there. For some reason the validation tests pass when I run

```
pytest tests/test_kernels_validation.py
```

but fail when I run the full test suite. The only failing test is `tests/test_kernels_validation.py::test_check_results[vector_grid_type_data_and_extra_kwargs-GridType.VECTOR_C_GRID]`. If I skip `test_viscosity_filter`, then the test works. This suggests that that test is somehow modifying the fixture. I'll try to get to the bottom of it.

tests/conftest.py
```diff
@@ -0,0 +1,215 @@
...
+tripolar_grids = [
+    GridType.TRIPOLAR_REGULAR_WITH_LAND_AREA_WEIGHTED,
+    GridType.TRIPOLAR_POP_WITH_LAND,
+]
+vector_grids = [GridType.VECTOR_C_GRID]
+
+
+def _make_random_data(shape: Tuple[int, int], seed: int) -> np.ndarray:
+    rng = Generator(PCG64(seed))
+    return rng.random(shape)
```
rabernat, line 67: Just noting a lesson learned: for the verification tests to work, the fixtures need to be deterministic. So here we are generating random data with a specific seed.

tests/conftest.py
```diff
@@ -0,0 +1,215 @@
...
+    grid_data[-1, (nx // 2) :] = half_northern_edge[::-1]
+    return grid_data
+
+
+@pytest.fixture(scope="session", params=scalar_grids)
+def scalar_grid_type_data_and_extra_kwargs(request):
+    grid_type = request.param
+    shape = (128, 256)
+
+    data = _make_random_data(shape, 100)
```
rabernat, line 103: ...and then we invoke the function with a fixed, constant seed, which is the same every time the test is run. That means we always get the same random data.

rabernat:
Ok, I think this is good to go.

tests/conftest.py
NoraLoose: Thanks for digging into this and getting to the bottom of the issue!

NoraLoose:
This looks great to me! The verification tests will be super valuable. And the entire test suite is cleaned up, too. Nice work.

I think we can merge this.

PR #80, opened 2021-08-02
Verification tests fixes

NoraLoose:
Here are some suggestions for how `add-verification-tests` in PR #79 has to be changed in order to get all tests to pass.

### Conservation test
The conservation test for `TRIPOLAR_POP_WITH_LAND` was failing because this Laplacian needs the northernmost row of `dxn` to fold onto itself. Same for `dyn`. This is an inherent property of the tripole grid, without which conservation cannot be satisfied. For example, in the graphic below the x-spacings of the northern grid edge cell of B and E have to be the same.
Screen Shot 2021-08-02 at 3 35 24 PM

I added the symmetry requirement of `dxn` and `dyn` to the fixture that creates the grid data. I also added a test that checks whether the user inputs grid data that satisfies this property.

*Side note for the curious*: This issue has not shown up before because we used identical grid data for `dxn` and `dyn`. If `dxn = dyn`, the flux across the northern edge simplifies to
```
nflux = ((np.roll(data, -1, axis=-2) - data) / self.dyn * self.dxn)
         = (np.roll(data, -1, axis=-2) - data),
```
so we accidentally cancelled out all problems that arose from non-symmetric `dxn` and `dyn` grid data.

### Flux test

@rabernat, I reverted the flux test back to your previous version. The test checks for *isotropic* diffusion, and therefore has to operate on regular grid data. 

Some more background on this test:
The "outlier" values (of 1000, 2000) for `dx`, `dy` are chosen far enough from the delta function so that `Laplacian(delta)` should not feel these outlier values when computing the fluxes, and isotropic diffusion will be satisfied. However, the outlier data is still close enough so that Laplacian(delta) will feel them if `np.roll(dx, -1, axis)` in `kernels.py` is mistakenly coded as `np.roll(dx, +1, axis)` and vice versa. I added these comments to the code, which will hopefully shed some light on how / why this test was designed. Sorry about the confusion!

### DIFF
--- a/gcm_filters/kernels.py
+++ b/gcm_filters/kernels.py
@@ -360,6 +360,21 @@ class POPTripolarLaplacianTpoint(BaseScalarLaplacian):
         if self.wet_mask[..., 0, :].any():
             raise AssertionError("Wet mask requires zeros in southernmost row")
 
+        # check that northern edge grid data folds onto itself
+        nx = np.shape(self.dxn)[-1]
+        first_half = self.dxn[..., [-1], : (nx // 2)]
+        second_half = self.dxn[..., [-1], (nx // 2) :]
+        if not np.all(first_half[..., ::-1] == second_half):
+            raise AssertionError(
+                "Northernmost row of dxn does not fold onto itself. This is a requirement for using a tripole boundary condition."
+            )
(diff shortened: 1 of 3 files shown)


rabernat:
Ok thanks! I'll merge this and keep working!

PR #82, opened 2021-08-24
Updating transition_width and n_steps

iangrooms:
This PR does 2 things:

1. Adds a default `n_steps` for general `transition_width`. Previously the default was used for any transition width, but was only accurate for transition widths >= pi.
2. Adds a test that checks if `transition_width > 1`. A user might think that the minimum transition width is 0, which is not true.

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -317,6 +317,10 @@ class Filter:
                     "dx_min must be set to 1."
                 )
 
+        # Check if transition_width is <=1
+        if self.transition_width <= 1:
+            raise ValueError(f"Transition width must be > 1.")
+
         # Get default number of steps
         filter_factor = self.filter_scale / self.dx_min
         if self.ndim > 2:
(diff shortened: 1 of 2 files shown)


gcm_filters/filter.py
```diff
@@ -282,7 +282,7 @@ class Filter:
         - ``FilterShape.TAPER``: The target filter has target grid scale Lf. Smaller scales are zeroed out.
           Scales larger than ``pi * filter_scale / 2`` are left as-is. In between is a smooth transition.
     transition_width : float, optional
-        Width of the transition region in the "Taper" filter.
+        Width of the transition region in the "Taper" filter. Theoretical minimumm is 1; not recommended.
```
NoraLoose, line 285: Small typo: minimumm  --> minimum

NoraLoose (approved):
Thanks @iangrooms for generalizing the scaling law for `n_steps_default` to general `transition_width` - really useful! The changes look great to me.

For the records: The new scaling law gives the same `n_steps_default` as the old scaling law for `transition_width` = np.pi.

Issue #89, opened 2021-09-14
Helper method for applying a filter to an xarray Dataset


asross:
I've found myself writing a wrapper method around `Filter#apply` that works for both `DataArray`s and `Dataset`s (applying to all variables that have the given `dims`):

```python
def apply(filter, ds, dims=['y','x']): 
    if isinstance(ds, xr.Dataset):
        filtered = ds.copy()
        for k, v in ds.variables.items():
            if all(d in v.dims for d in dims):
                filtered[k] = filter.apply(ds[k], dims=dims)
        return filtered
    else:
        return filter.apply(ds, dims=dims)
```

Would it make sense to make a PR that wraps the [current definition of `Filter#apply`](https://github.com/ocean-eddy-cpt/gcm-filters/blob/master/gcm_filters/filter.py#L414) with a helper of this kind?

rabernat:
Yes, I think this would be a very useful contribution. Thanks a lot for your offer!

Please make sure to include tests for any new features you propose in your PR.

asross:
Awesome! I created a PR :)

PR #90, opened 2021-09-14
Support application of filters to xarray datasets

asross:
Would resolve https://github.com/ocean-eddy-cpt/gcm-filters/issues/89.

Note that some of the tests on `master` appear to be failing for me (on Python 3.9.5) -- but this change doesn't cause any new tests to fail for me, and the test I added passes.

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -411,7 +411,19 @@ class Filter:
         ax.grid(True)
         ax.legend()
 
-    def apply(self, field, dims):
+    def apply(self, field_or_dataset, dims):
+        """Filter a field or xarray dataset with scalar Laplacian across the
+        dimensions specified by dims."""
+        if isinstance(field_or_dataset, xr.Dataset):
+            filtered = field_or_dataset.copy(deep=True)
+            for key, field in filtered.variables.items():
+                if all(dim in field.dims for dim in dims):
+                    filtered[key] = self.apply_to_field(field, dims=dims)
+            return filtered
+        else:
+            return self.apply_to_field(field_or_dataset, dims=dims)
+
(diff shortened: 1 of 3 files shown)


gcm_filters/filter.py
```diff
@@ -411,7 +411,19 @@ def plot_shape(self, ax=None):
...
 
-    def apply(self, field, dims):
+    def apply(self, field_or_dataset, dims):
+        """Filter a field or xarray dataset with scalar Laplacian across the
+        dimensions specified by dims."""
+        if isinstance(field_or_dataset, xr.Dataset):
+            filtered = field_or_dataset.copy(deep=True)
+            for key, field in filtered.variables.items():
+                if all(dim in field.dims for dim in dims):
+                    filtered[key] = self.apply_to_field(field, dims=dims)
```
NoraLoose, line 421: Should we add a warning here if `field` does not have the correct `dims`? (Just a suggestion - I'm not feeling strongly about it.) Otherwise, there is no obvious way for the user to see if fields in the dataset got actually filtered or not. For instance, the user may expect all fields in the dataset to be filtered but forgot to rename the spatial dimensions to `dims` for some of the fields. 

For context, if the user applies the `apply` method to a field (rather than a dataset) that does not have the dimensions `dims`, then an error message will be raised.

asross: (Added a commit to that effect, with some tests!)

NoraLoose: Great, I think the added warning is useful. Thanks for your work!

NoraLoose:
This looks great to me and ready to be merged! Let's wait a few more days in case anyone else has comments.

PR #109, opened 2021-12-10
Remove notebook directory with packaging demo 

NoraLoose:
This PR is in reply to @callumrollo's [comment](https://github.com/openjournals/joss-reviews/issues/3947#issuecomment-990587039) on the JOSS review page, which I am reposting here:

> There is a notebooks folder in the top level of the repo, with a jupyter notebook enticingly named tutorial.ipynb. However, these appears to be a demo for a lesson on packing or similar? Not sure what it's doing here. I'd suggest removing it lest potential users mistake it for a demo of the package functionality.
> 
> The documentation contains some great examples of package functionality. Perhaps these could be dumped into a notebook for ease of experimentation by potential users?

Thanks for the catch, Callum! I suspect that the `notebook` directory is a remnant of copying over the repo structure from somewhere else. So let's delete it to avoid confusion among users.

I'm hesitant to copy over our examples from `docs/examples` to `notebooks` since this would lead to redundant code that has to be maintained separately. Any other opinions on that?


rabernat:
> I'm hesitant to copy over our examples from `docs/examples` to `notebooks` since this would lead to redundant code that has to be maintained separately. Any other opinions on that?

Why can't we just direct users directly to the `docs/examples` notebooks? We could even create a binder link on the repo that would open these notebooks directly.

NoraLoose:
> We could even create a binder link on the repo that would open these notebooks directly.

Sounds good, @rabernat. Let's open a separate PR for that?

Issue #118, opened 2021-12-11
add environment file/ extra dependencies for example notebooks


isgiddy:
I was running through the documentation file examples - which are really great! and needed to install dependencies along the way. I noticed there was an environment file in the docs folder but think this is not for running the individual notebooks, but rather compiling the documentation?

For users who might want to do what I did, it could be nice to state these extra dependencies (beyond numpy, xarray and dask) /provide an environment file to run through the examples.

Listing them here for reference:
matplotlib
cupy (optional)
intake
intake-xarray
s3fs
cartopy

NoraLoose:
Thanks for pointing out this issue, @isgiddy! I just opened a PR (#120) that should solve this for the user.

I'm also tagging our [JOSS review thread](https://github.com/openjournals/joss-reviews/issues/3947) so we can keep track over there of which issues are open / closed.

PR #120, opened 2021-12-14
Optional dependencies to run notebooks

NoraLoose:
Handles #118.


NoraLoose:
There seems to be a problem with conda-forge and either of the new packages added to `docs/environment.yml`. The same problem will arise when users try to install `gcm-filters[examples]` (see new `[options.extras_require]` section in `setup.cfg`). Converting this PR to draft again - not sure how to proceed, though.

NoraLoose:
@jbusecke maybe you know how to proceed with this issue?

README.md
```diff
@@ -20,15 +20,22 @@ Through integration with [dask](https://dask.org/), GCM-Filters enables parallel
...
 ### Installation
 
-GCM-Filters can be installed with pip:
+GCM-Filters can be installed using conda:
+```shell
+conda install -c conda-forge gcm_filters
+```
+If you want to run the example notebooks in the [GCM-Filters documentation](https://gcm-filters.readthedocs.io/), you will need a few extra dependencies that you can install via:
+```shell
+conda install -c conda-forge gcm_filters[examples]
```
jbusecke, line 29: I have to admit, I am not even sure if conda allows this syntax? I have only ever seen these 'variants' with pip.

jbusecke:
Unless conda does support this sort of syntax (a very quick search was unsuccessful for me) I think there are two options here:
1. Add a specific`environment_examples.yml` file with everything needed to run
2. Include the additional dependencies in the doc environment. 

Generally I think 2 is nicer, if RTD can handle the additional install without too much delay/issues.

jbusecke:
Ah sorry, I did not realize that the RTD build was indeed failing....

jbusecke:
How about you try to use [mamba instead of conda](https://docs.readthedocs.io/en/stable/guides/build-using-too-many-resources.html#use-mamba-instead-of-conda)?

jbusecke:
I am trying the switch over at https://github.com/xgcm/xgcm/pull/401. Once that works, Id be happy to implement this here.

NoraLoose:
I ended up including the additional dependencies that are necessary to run the notebooks in `docs/environment.yml`. This seemed like the most straight-forward thing to do. 

Maybe someone could look over this PR? It's the last issue that has to be addressed to get our package published in JOSS. 🎉

jbusecke:
Thats usually my approach too. Looking at it now.

docs/installation.rst
```diff
@@ -46,3 +48,13 @@ GCM-Filters and submit issues_ and `pull requests`_.
...
 .. _issues: https://github.com/ocean-eddy-cpt/gcm-filters/issues
 .. _`pull requests`: https://github.com/ocean-eddy-cpt/gcm-filters/pulls
+
+
+How to run the example notebooks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you want to run the example notebooks in this documentation, you will need a few extra dependencies that you can install via::
+
+   mamba env create -f docs/environment.yml
```
jbusecke, line 58: I just noticed that we suggest conda elsewhere. It might be better to do all examples with conda and make a note somewhere that mamba can be used as a faster drop in replacement?

jbusecke, line 58: I just noticed that we suggest conda elsewhere. It might be better to do all examples with conda and make a note somewhere that mamba can be used as a faster drop in replacement?

jbusecke (approved):
Looks good to me. Thanks a lot for the work!

NoraLoose:
Awesome, thanks @jbusecke! I agree that we should be consistent with conda vs. mamba. In the installation instructions (`installation.rst`), we just talk about conda now.

jbusecke (approved):
Looks good to me. Shall we merge this?

PR #123, opened 2021-12-18
Fix check for tripolar Laplacian

### LINKED ISSUE #122: Tripole boundary check does not pass on POP data
While re-running the [POP tripolar grid example](https://gcm-filters.readthedocs.io/en/latest/examples/example_tripole_grid.html) (while working on PR #115) I noticed that the following "tripole" check does not pass for the POP grid data

https://github.com/ocean-eddy-cpt/gcm-filters/blob/7898c9cab7463de5d86bac95958239c0413a0a79/gcm_filters/kernels.py#L502-L509

As a result, running the [tripolar grid notebook in the documentation](https://gcm-filters.readthedocs.io/en/latest/examples/example_tripole_grid.html) gives an AssertionError. It seems like we haven't re-run that notebook since the check above was added.

I will try to get to the bottom of this later today.

NoraLoose:
Closes #122.

### Problem description

The tripolar Laplacian `POPTripolarLaplacianTpoint` checks that the northern edge grid data folds onto itself - a necessary criterion of the tripole grid geometry. However, this check failed when this Laplacian got applied to some *actual* POP grid data (rather than just our synthetic test data in our package), such as in our [example notebook in the documentation](https://gcm-filters.readthedocs.io/en/latest/examples/example_tripole_grid.html).

### Solution

I changed the check to only test along northern edge grid data **that is not on land**. The POP grid data goes a little bit crazy on land points--and does whatever--, so we don't want to check on land. The check now passes.

### Lesson learned 

This brings us back to an old question in #5: Should we do our tests on real model data, or at least, real model *grid* data (rather than synthetic data)? We would have discovered this issue much earlier...

### DIFF
--- a/gcm_filters/kernels.py
+++ b/gcm_filters/kernels.py
@@ -524,8 +524,8 @@ class POPTripolarLaplacianTpoint(BaseScalarLaplacian):
         # note: grid data goes crazy for POP model land points so we don't want to check for land points
         nx = np.shape(self.dxn)[-1]  # number of longitudes or columns
         # grab second to last row since we have already appended one extra row
-        first_half = np.where(self.n_wet_mask == 1, self.dxn, 0)[...,-2, :(nx // 2)]
-        second_half = np.where(self.n_wet_mask == 1, self.dxn, 0)[...,-2, (nx // 2):]
+        first_half = np.where(self.n_wet_mask == 1, self.dxn, 0)[..., -2, : (nx // 2)]
+        second_half = np.where(self.n_wet_mask == 1, self.dxn, 0)[..., -2, (nx // 2) :]
         if not np.all(first_half[..., ::-1] == second_half):
             raise AssertionError(
                 "Northernmost row of dxn does not fold onto itself. This is a requirement for using a tripole boundary condition."
--- a/gcm_filters/kernels.py
+++ b/gcm_filters/kernels.py
@@ -499,27 +499,16 @@ class POPTripolarLaplacianTpoint(BaseScalarLaplacian):
         if self.wet_mask[..., 0, :].any():
             raise AssertionError("Wet mask requires zeros in southernmost row")
 
-        # check that northern edge grid data folds onto itself
-        nx = np.shape(self.dxn)[-1]
-        first_half = self.dxn[..., [-1], : (nx // 2)]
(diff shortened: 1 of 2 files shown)


rabernat:
Hi Nora! Thanks for working on this important issue. Unfortunately I am about to unplug for my holiday break. I will be back online the first week of January and will try to review this then. If you want to move faster, then you don't have to wait for me. On the other hand, I hope you (and everyone else involved) is also planning to take a break as well! 🎄🎅🎄

NoraLoose:
Yes, let's finish this PR and the remaining issues related to the JOSS review in the first week of January. Happy holidays!

NoraLoose:
Would anyone be willing to review this PR?

gcm_filters/kernels.py
```diff
@@ -531,6 +520,25 @@ def __post_init__(self):
         # note: wet_mask(j+1,i) corresponds to np.roll(wet_mask, -1, axis=-2)
         self.n_wet_mask = self.wet_mask * np.roll(self.wet_mask, -1, axis=-2)
 
+        # check that northern edge grid data folds onto itself if not on land;
+        # note: grid data goes crazy for POP model land points so we don't want to check for land points
+        nx = np.shape(self.dxn)[-1]  # number of longitudes or columns
+        # grab second to last row since we have already appended one extra row
+        first_half = (self.n_wet_mask * self.dxn)[..., [-2], : (nx // 2)]
```
iangrooms, line 527: It might be better to do this test with boolean indexing rather than multiplying by the wet mask. If, for example, some of the land data is `NaN` then I think this test would throw an error because `0 * np.NaN` is `np.NaN` and `np.NaN == np.NaN` returns `False`. Maybe something like
`first_half = self.dxn[...,-2,self.n_wet_mask[...,-2,:(nx//2)] == 1]`
would avoid NaN problems.

NoraLoose: Thanks for the catch! In my latest commit I followed your suggestion - just using a different syntax that will also work with dask and cupy.

NoraLoose:
@iangrooms do you now approve this PR? If so I would like to merge this, so we can move forward with our JOSS revisions.

iangrooms (approved):
Looks good to me. Thanks for  your work on this!

PR #115, opened 2021-12-10
Use Pooch for downloads

rabernat:
This PR updates the example notebooks to use [Pooch](https://www.fatiando.org/pooch/latest/index.html) for downloading the data. This is better than requests because it is
- more transparent where the data are coming from (download via DOI)
- uses memory more efficiently (streams the file to disk)
- includes md5 checksum verification

PR #127, opened 2022-01-10
Factor Gaussian to improve stability

iangrooms:
This PR gives users the option of splitting a Gaussian filter into a sequence of Gaussian filters each of which has a smaller length scale. Each smaller-scale Gaussian filter is more stable to roundoff. For the true/target filter, and in the absence of roundoff errors, there is no difference between applying one Gaussian filter or applying several small-scale filters. Even without roundoff errors, it is not _exactly_ equivalent to filter once versus more than once because of the polynomial approximation to the Gaussian filter. But because our defaults give a very high accuracy in the polynomial approximation, the two methods give extremely similar results, and the factored method can be more stable to roundoff. 

This PR is marked as Draft because @NoraLoose is going to this method it to the tutorial on numerical stability.


NoraLoose:
Thanks @iangrooms for implementing this. I think it may be better if I put in a separate PR for showing this feature in the [numerical instability notebook](https://gcm-filters.readthedocs.io/en/latest/examples/example_numerical_instability.html). I would like to avoid conflicts with PR #115, which also changes the same notebook in the documentation. The latter PR is essentially ready to merge but needs PR #123 to be reviewed and merged first. So let's not complicate things further.

If you change this PR from "Draft" to "Ready for review", I can have a look.

docs/theory.rst
```diff
@@ -200,9 +200,10 @@ The usual manifestation of these roundoff errors is high-amplitude small-scale n
...
-    4. The final option is simply to use a different approach to filtering, not based on ``gcm-filters``.
+    2. Applying a Gaussian filter with ``filter_scale`` :math:`L` twice is the same as applying a Gaussian filter with ``filter_scale`` :math:`\sqrt{2} L` once. More generally, one can break a single large-scale Gaussian filter up into several Gaussian filters each of which has a smaller filter scale. The Filter class has an argument ``n_iterations`` that will automatically split a Gaussian filter into a sequence of smaller filters, each of which is less sensitive to roundoff errors. If a user is encountering instability with the standard Gaussian filter, the user can try setting ``n_iterations`` to an integer greater than 1. The user should input their desired filter scale, and then the code will automatically adjust it to reflect the scale of the smaller constituent filters. Applying the Taper filter twice is not equivalent to applying it once with a different scale, so this method only works for the Gaussian filter.
```
NoraLoose, line 203: suggested syntax change: Filter class --> :py:class:`gcm_filters.Filter` class

"... that will automatically split a Gaussian filter into a sequence of `n_iterations` smaller filters, ..."

gcm_filters/filter.py
```diff
@@ -317,10 +327,23 @@ def __post_init__(self):
                     "dx_min must be set to 1."
                 )
 
+        # Check for n_iterations is < 1
+        if self.n_iterations < 1:
+            raise ValueError(
+                f"Number of filters into which the Gaussian is factored must be >= 1."
```
NoraLoose, line 333: This error message could be confusing if the user has specified a Taper filter shape.

docs/theory.rst
NoraLoose: We could be even more specific and replace the sentence

"Applying a Gaussian filter with ``filter_scale`` :math:`L` twice is the same as applying a Gaussian filter with ``filter_scale`` :math:`\sqrt{2} L` once."

by

"Applying a Gaussian filter with ``filter_scale`` :math:`L` `N` times is the same as applying a Gaussian filter with ``filter_scale`` :math:`\sqrt{N} L` once."

NoraLoose:
Thanks @iangrooms for this new feature. The option to break large filters into a sequence of smaller filters is really exciting because this may bail `gcm-filters` out of the numerical instability issue. 🎉 

I left some minor comments above. A major comment is that we should add a test in `filter.py` that checks on some test data that specifying `filter_scale=L` and `n_iterations=1` gives indeed (approximately) the same as specifying `filter_scale=L/sqrt(2)` and `n_iterations=2`.

rabernat:
This is awesome! Thanks so much Ian. Sounds like this approach could really help out with numerical stability. I agree with Nora that tests are a must.

I wonder if there is a generalization of this splitting approach that might also work with the taper filter.

docs/theory.rst
iangrooms: I'm updating the syntax. For the second suggestion, I kind of like it the way it is. If I updated the sentence a suggested then I would need to delete the next sentence, which basically says the same thing in words. Also, the proof of the `N` step case is just a repeated application of the two-step, so starting directly from `N` sort of hides how the proof works. Not that it will matter to most readers. I can change it if you (or other reviewers) feel strongly about it.

gcm_filters/filter.py
iangrooms: OK, I will drop 'Gaussian.' If they are using the Taper filter and setting `n_iterations` to anything other than 1, then it will raise a warning later.

docs/theory.rst
NoraLoose: Independent of whether we write about the two-step or `N`-step case in the `theory.rst`, would it make sense to have the "proof" in the documentation? (I think it's a one-liner, right?) Maybe we could outsource it to the [numerical instability example notebook](https://gcm-filters.readthedocs.io/en/latest/examples/example_numerical_instability.html). I anticipate users asking about it because the splitting is not described in the Grooms et al. 2021 paper. It would also clarify why it doesn't work for the Taper.

iangrooms: It would have been super easy to add to the JAMES paper, if only we'd thought of it... Adding it to the theory.rst would take some space: it's a one-line proof if all the notation is defined, but we don't have the notation defined yet. 

When I was cooking up a test to compare the iterated and original filters, I derived an error bound. (the new test just checks that the error bound is satisfied.) I could add a new "chapter" to the documentation (not a new section of theory.rst) that proves the equivalence, derives the error bound, and explains why it won't work for the Taper filter? Not sure the best way to do this.

iangrooms:
Replying to @rabernat's comment about doing something like this for the Taper filter. I wonder too, but haven't had any good ideas yet. We could apply a sequence of Taper filters that successively zero out more and more small scales. This might improve the stability, because we know that filtering smooth fields tends to be more stable than filtering rough fields. But it also might not help, because the last filter you apply still has a big filter "factor" and could be unstable. The nice thing about the Gaussian version is that you get away with only ever using filters with small filter scales, and I can't figure out how to do that with the Taper.

docs/theory.rst
NoraLoose: Maybe a new subsection of theory.rst would be best. Are you worried it will get too long?

tests/test_filter.py
```diff
@@ -383,6 +405,24 @@ def test_application_to_dataset():
...
+    # The following tests whether a relative error bound in L^2 holds.
+    # If the polynomial approximations are accurate to within 0.01,
+    # then the precise bound would have 0.0301 ** 2 instead of 0.04 ** 2.
+    # Bumped up because the default n_steps is not guaranteed to give
+    # error less than *exactly* 0.01; just quite close.
+    assert ((filtered_dataset.spatial - filtered_dataset_n2.spatial) ** 2).sum() < (
+        0.04 ** 2
+    ) * (filtered_dataset.spatial ** 2).sum()
+
     # Temporal variables should be unaffected because the filter is only
```
NoraLoose, line 426: Thanks for adding this test for filter splitting! You put the test inside of `test_application_to_dataset()`, where the latter test was specifically designed to check if feature #90 does the correct thing, i.e., whether the filter works on a`xarray.Dataset` (rather than just an `xarray.DataArray`). 

It would be cleaner to introduce a separate test for the filter splitting and then run the test on all implemented grids / Laplacians. Something like

```
def test_filter_splitting(grid_type_and_input_ds, filter_args):
    "Some description of the test"

    grid_type, da, grid_vars = grid_type_and_input_ds

    iterated_filter_args = filter_args.copy()
    iterated_filter_args["n_iterations"] = 2

    filter = Filter(grid_type=grid_type, grid_vars=grid_vars, **filter_args)
    iterated_filter = Filter(grid_type=grid_type, grid_vars=grid_vars, **iterated_filter_args)

    filtered = filter.apply(da, dims=["y", "x"])
    iteratively_filtered = iterated_filter.apply(da, dims=["y", "x"])

    # now your assert statement

```

iangrooms: I updated the tests. I am a rank amateur at Python testing, so please don't hesitate to criticize this.

docs/theory.rst
iangrooms: In the latest commit I added a new section to theory.rst. It's a bit long; let me know what you think.

docs/theory.rst
```diff
@@ -189,6 +189,79 @@ The example below shows what happens with ``n_steps`` = 4.
...
+
+.. math:: g_L(k) = p_1(k^2) + e_1(k)
+
+and for :math:`p_N` we have the error
+
+.. math:: g_{L/\sqrt{N}}(k) = p_N(k^2) + e_N(k)
+
+The aforementioned property of the Gaussian implies that
+
+.. math:: g_{L}(k) = (p_N(k)^2 + e_N(k))^N = (p_N(k^2))^N + N (p_N(k^2))^{N-1} e_N(k) + \ldots + (e_N(k))^N
```
NoraLoose, line 235: p_N(k)^2 --> p_N(k^2)

docs/theory.rst
```diff
@@ -189,6 +189,79 @@ The example below shows what happens with ``n_steps`` = 4.
...
+
+(using the binomial expansion.)
+Subtracting gives us an expression for the difference between the polynomial approximation with scale :math:`L` and the factored approximation using :math:`N` filters each with scale :math:`L/\sqrt{N}`:
```
NoraLoose, line 238: This is optional, but rather than saying "subtracting" we could say "subtracting (1) from (2)" via equation cross-references and labels, see [here](https://developer.lsst.io/v/DM-5063/docs/rst_styleguide.html#math).

iangrooms: Good idea, and wasn't too hard to implement.

tests/test_filter.py
```diff
@@ -408,6 +430,50 @@ def test_application_to_dataset():
...
+
+    # The following tests whether a relative error bound in L^2 holds.
+    # If the polynomial approximations are accurate to within 0.01,
+    # then the precise bound would have 0.0301 ** 2 instead of 0.04 ** 2.
+    # Bumped up because the default n_steps is not guaranteed to give
+    # error less than *exactly* 0.01; just quite close.
+    assert (((filtered - iteratively_filtered) ** 2) * area).sum() < (0.04 ** 2) * (
+        (filtered ** 2) * area
+    ).sum()
+
```
NoraLoose, line 475: More generally, we could test for
```
assert (((filtered - iteratively_filtered) ** 2) * area).sum() < ((iterated_filter_args["n_iterations"]+1) ** 2) * (
        (filtered ** 2) * area
    ).sum()
```
In the comment above this assertion statement, we could refer to the error bound derived in `factored_gaussian.rst` so people will understand this test.

tests/test_filter.py
```diff
@@ -408,6 +430,50 @@ def test_application_to_dataset():
...
+        )
+    ],
+)
+def test_iterated_filter(grid_type_and_input_ds, filter_args):
+    "Test that the iterated Gaussian filter gives a result close to the original"
+
+    grid_type, da, grid_vars = grid_type_and_input_ds
+
+    iterated_filter_args = filter_args.copy()
+    iterated_filter_args["n_iterations"] = 2
```
NoraLoose, line 451: Should we add another iterated filter with `n_iterations > 2`? This is optional.

docs/factored_gaussian.rst
```diff
@@ -0,0 +1,94 @@
...
+Applying the Taper filter repeatedly is not equivalent to applying it once with a different scale, so this method only works for the Gaussian filter.
+
+.. note:: When ``n_iterations`` is greater than one, ``n_steps`` is the number of steps in a single small-scale Gaussian filter. The total number of steps taken by the algorithm is the product of ``n_iterations`` and ``n_steps``. If ``n_steps`` is not set by the user, or if it is set to a value less than 3, the code automatically changes ``n_steps`` to a default value that gives an accurate approximation to the small-scale Gaussian filters that comprise the factored filter.
```
NoraLoose, line 10: super small suggested change: "greater than one" --> "greater than 1"

NoraLoose (commented):
This is great!! The new documentation section is written very clearly, and the new code and test looks good to me.

Final comment / question is whether we want to implement the same test for the viscosity filter, too (for completeness).

iangrooms:
> This is great!! The new documentation section is written very clearly, and the new code and test looks good to me.
> 
> Final comment / question is whether we want to implement the same test for the viscosity filter, too (for completeness).

The latest commit adds a test of the iterated Gaussian for the vector filter. It uses `n_iterations = 4` so it satisfies @NoraLoose's other suggestion too, as well as the minor changes to comments and docs.

tests/test_filter.py
```diff
@@ -441,3 +508,42 @@ def test_viscosity_filter(vector_grid_type_and_input_ds, filter_args):
...
+
+
+@pytest.mark.parametrize(
+    "filter_args",
+    [dict(filter_scale=4.0, dx_min=1.0, n_steps=0, filter_shape=FilterShape.GAUSSIAN)],
+)
+def test_iterated_viscosity_filter(vector_grid_type_and_input_ds, filter_args):
+    """Test error in the iterated Gaussian filter for vectors"""
+    grid_type, da_u, da_v, grid_vars, geolat_u = vector_grid_type_and_input_ds
+
```
NoraLoose, line 520: suggested change:
```
grid_type, da_u, da_v, grid_vars, _ = vector_grid_type_and_input_ds
```

tests/test_filter.py
```diff
@@ -408,6 +430,51 @@ def test_application_to_dataset():
...
+    area = 1
+    for k, v in grid_vars.items():
+        if "area" in k:
+            area = v
+            break
+
+    # The following tests whether a relative error bound in L^2 holds.
+    # If the polynomial approximations are accurate to within 0.01,
+    # then the precise bound would have 0.0301 ** 2 instead of 0.04 ** 2.
+    # See the "Factoring the Gaussian Filter" section of the docs for details.
```
NoraLoose, line 470: Is `0.0301 ** 2` a remnant of the old error bound? Following the "Factoring the Gaussian Filter" section, I get `0.0001 * (1+2) ** 2 = .0009` (not `0.0301 ** 2`)

tests/test_filter.py
```diff
@@ -408,6 +430,51 @@ def test_application_to_dataset():
...
+    # Bumped up the tolerance because the default n_steps is not
+    # guaranteed to give error less than *exactly* 0.01; just quite close.
+    assert (((filtered - iteratively_filtered) ** 2) * area).sum() < (0.04 ** 2) * (
+        (filtered ** 2) * area
+    ).sum()
+
```
NoraLoose, line 476: In `test_iterated_viscosity_filter`, we test against `error_bound * unfiltered` (rather than `error_bound * filtered`). Maybe good to be consistent between the two tests?

NoraLoose (commented):
Thanks for adding the test for the viscosity filter! Looks great - just some minor suggestions / questions.

tests/test_filter.py
iangrooms: The bound in the docs is asymptotic; it ignores all terms `0.01 ** p` with exponent higher than 2. I did that because it simplifies the bound when there are lots of filter steps. For this specific test `n_iterations = 2` so I don't need an asymptotic approximation.

NoraLoose: Great, makes sense! Let's include this explanation in the comment so we remember that the case `n_iterations = 2` is "special" and doesn't need the asymptotic approximation from the docs.

tests/test_filter.py
iangrooms: This was a mistake; it should compare to unfiltered, per the docs. Will fix in the next commit.

tests/test_filter.py
iangrooms: Will update in the next commit.

NoraLoose:
The PR looks great to me. I don't have any further comments. 

It would be nice to have another reviewer look over this PR before we merge, since this PR makes some major changes to the code. Any volunteers?

docs/factored_gaussian.rst
```diff
@@ -0,0 +1,94 @@
+Factoring the Gaussian Filter
+=============================
+
+This section provides background on applying a Gaussian filter with a large filter scale by repeatedly applying Gaussian filters with a smaller filter scale.
+This was not discussed in `Grooms et al. (2021) <https://doi.org/10.1029/2021MS002552>`_, so this section provides extra detail.
+
+The :py:class:`gcm_filters.Filter` class has an argument ``n_iterations`` that will automatically split a Gaussian filter into a sequence of Gaussian filters with smaller scales, each of which is less sensitive to roundoff errors. If a user is encountering instability with the standard Gaussian filter, the user can try setting ``n_iterations`` to an integer greater than 1. In general ``n_iterations`` should be set to the smallest value that avoids numerical instability. The user should input their desired filter scale, and then the code will automatically adjust it to reflect the filter scale of the smaller constituent filters.
```
rabernat, line 7: In the sentence above - "adjust it" - I can't understand what **it** refers to. In general this section should try to be clearer and more explicit about what exactly is the recommended procedure for users.

docs/factored_gaussian.rst
```diff
@@ -0,0 +1,94 @@
...
+Applying the Taper filter repeatedly is not equivalent to applying it once with a different scale, so this method only works for the Gaussian filter.
+
+.. note:: When ``n_iterations`` is greater than 1, ``n_steps`` is the number of steps in a single small-scale Gaussian filter. The total number of steps taken by the algorithm is the product of ``n_iterations`` and ``n_steps``. If ``n_steps`` is not set by the user, or if it is set to a value less than 3, the code automatically changes ``n_steps`` to a default value that gives an accurate approximation to the small-scale Gaussian filters that comprise the factored filter.
```
rabernat, line 10: It does not make sense to me that we would silently override an explicit value of `nsteps=3`. If this is not a valid choice, we should raise a `ValueError`. There should be one and only one way to trigger the automatic determination of `n_steps`, e.g. `n_steps=None`.

docs/theory.rst
```diff
@@ -200,9 +200,10 @@ The usual manifestation of these roundoff errors is high-amplitude small-scale n
     In such cases, the user has a few options to try to regain stability.
 
     1. If the data being filtered is single-precision, it might help to promote it to double precision (or higher) before filtering.
-    2. The user can also try reducing ``n_steps``, but must not reduce it too much or the resulting filter will not behave as expected.
-    3. Users might elect to *coarsen* their data before filtering, i.e. to reduce the resolution of the input data before applying the filter. This has the effect of increasing the grid size, and thus decreasing the gap between the filter scale and the grid scale.
-    4. The final option is simply to use a different approach to filtering, not based on ``gcm-filters``.
+    2. If a user is encountering instability with the standard Gaussian filter, the user can try setting ``n_iterations`` to an integer greater than 1. Read more about this in :doc:`factored_gaussian`.
+    3. The user can also try reducing ``n_steps``, but must not reduce it too much or the resulting filter will not behave as expected.
```
rabernat, line 204: Can we give better guidance on what is meant by "too much"?

tests/test_filter.py
rabernat: To test more iterations, we easily could add an additional `@pytest.mark.paratmetrize('n_iterations', [2, 3, 4]))` decorator to the function. This seems like a good idea to me.

rabernat (commented):
Looks really good. I have a a few minor suggestions on documentation wording and test strategy.

docs/factored_gaussian.rst
iangrooms: This isn't really a comment about this PR per se; I'm just describing how the current code treats `n_steps`. Could we deal with this separately?

docs/theory.rst
iangrooms: This isn't really a comment about this PR; the offending snippet is in the main branch. Could we deal with this separately?

tests/test_filter.py
```diff
@@ -441,3 +508,46 @@ def test_viscosity_filter(vector_grid_type_and_input_ds, filter_args):
...
+
+@pytest.mark.parametrize(
+    "filter_args",
+    [dict(filter_scale=4.0, dx_min=1.0, n_steps=0, filter_shape=FilterShape.GAUSSIAN)],
+)
+@pytest.mark.parametrize(
+    "n_iterations",
+    [2, 3, 4],
+)
```
NoraLoose, line 520: Do you need to re-parameterize? I think it will work if you delete these 4 lines because they are just repeated from above.

iangrooms: I deleted the lines (the repeated ones, not the first ones) and it threw an error, so I guess I need the lines. I don't really know much about how `pytest` works.

NoraLoose: Oh ok, my bad! I'm still learning about `pytest` too, so now we know one more thing about it. 😃

PR #130, opened 2022-02-01
Factor gaussian example

iangrooms:
@NoraLoose was going to update several things about the examples, so we planned to have her update the numerical stability example notebook with the new factored Gaussian filter. Plans changed, so I put together this PR to add an example using the factored Gaussian to the numerical instability notebook. While doing it I made several minor updates to the example notebook, including a new comparison at the end between the results of filtering with double precision, with the factored Gaussian, and via coarsening. (Edit:) This is basically finishing off #127.


rabernat:
This looks great Ian! I just read through the updated notebook and didn't catch any obvious issues. So I'll merge.

PR #131, opened 2022-02-02
Add warnings about dimension order to docs

### LINKED ISSUE #125: Order of filter dimensions matters
For most of our filter kernels, it matters whether the user does
```
filtered = filter.apply(field, dims=['y',  'x'])  # correct filtering
```
versus
```
filtered = filter.apply(field, dims=['x',  'y'])  # incorrect filtering
```

This came up in [this issue](https://github.com/ocean-eddy-cpt/gcm-filters/issues/124#issuecomment-998221195) in the context of filtering on a tripole grid. Our code needs the user to specify the correct order `['y',  'x']` because the tripolar exchanges are coded by assuming that `y` (latitude) is the second to last dimension:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/75f172c7c74c1f2f3b91e36f7345bd138f01c881/gcm_filters/kernels.py#L32-L39

Similarly, other Laplacians compute wet masks for the southern (and western) cell edges:

https://github.com/ocean-eddy-cpt/gcm-filters/blob/75f172c7c74c1f2f3b91e36f7345bd138f01c881/gcm_filters/kernels.py#L279-L281

This also assumes that latitude is the second to last dimension.

We should do either of the following two things:

1.  Change the code to the effect that it can deal with an incorrect dimension order `['x', 'y']`.
2. Communicate more clearly in the API and documentation that the dimension order matters.

rabernat:
Closes #125

Also includes general improvements to the docstrings of `apply` and `apply_to_vector` to match numpydoc conventions.


NoraLoose:
Thanks so much for working on this @rabernat! The added warnings and new docstrings are great. I will merge this.

PR #134, opened 2022-02-07
Factoring the Gaussian - Part 3

NoraLoose:
This PR makes a few minor tweaks to code and documentation related to factoring the Gaussian.

* In the Filter Class, distinguish between `self.filter_scale` and `self.filter_scale_iterated`. As a result, the string representation of the filter can give back the original filter scale (`self.filter_scale`) that the user put in, rather than the smaller filter scale associated with the constituent filters. I think this makes it more intuitive.
* I added a few code snippets to `factored_gaussian.rst` to exemplify the factoring.
* I added a few clarifying sentences and comments to the numerical instability example notebook.


gcm_filters/filter.py
```diff
@@ -421,14 +423,14 @@ def plot_shape(self, ax=None):
             linewidth=4,
         )
         ax.axvline(
-            2 * np.pi / self.filter_scale,
+            2 * np.pi / self.filter_scale_iterated,
```
iangrooms, line 426: On the lines above, the target that is plotted is `F(x) ** self.n_iterations` and the approximation that is plotted is `np.polynomial.chebyshev.chebval(x, self.filter_spec.p) ** self.n_iterations`. For these plots I think we want the vertical line to be at `filter_scale` not `filter_scale_iterated` right?

iangrooms:
OK, looks good to me. Thanks @NoraLoose!

NoraLoose:
Thanks Ian!

Issue #136, opened 2022-02-07
[BUG] Inconsistent dimensionality of discrete Laplacians


iangrooms:
Prompted by #106, I started to add some information about units to the code comments and documentation. In doing so I realized that there is a potential problem for users using either `REGULAR` or `REGULAR_WITH_LAND` grid types. Suppose a user has a regular grid with spacing 5 km, and wants to filter to a scale of 40 km. It would be completely reasonable for them to input `dx_min = 5000` and `filter_scale = 40000`, but that would produce wrong results. 

The problem is a dimensional inconsistency in the filter steps. A Laplacian filter step has the form `field_bar += (1 / s_l) * Laplacian(field_bar)`. In order for this to be dimensionally consistent, the dimensions of the Laplacian operator have to be the same as the dimensions of `s_l`. (Similar for biharmonic steps and `s_b`.) The way the code works right now, the dimensions of `s_l` are the same as the dimensions of `dx_min ** (-2)`, but the `REGULAR` Laplacians (including the `AREA_WEIGHTED` ones) are nondimensional because they implicitly use a grid size of 1.

I can think of a few ways to fix this to avoid potential future problems.
1. Force the user to set `dx_min = 1` for _all_ the `REGULAR` grid types. The `AREA_WEIGHTED` grid types already enforce dimensional consistency by throwing an error if `dx_min` is not 1. While it would work, it seems inelegant, and I'm not sure how to implement it anyways. 
2. Make all the `REGULAR` Laplacians dimensional by simply dividing the result produced by the existing code by `dx_min ** 2`. I like this because the variable names `dx_min` and `filter_scale` mean exactly what they say, and you can use any units you want (cm, m, km, whatever) as long as they're consistent. With this approach you could even set `dx_min = 1` and then let `filter_scale` be the nondimensional filter factor and it would _still_ work. I'm not sure how to effectively implement this though.
2. Make _everything_ nondimensional. To accomplish this we would set `dx_min=1` (no longer a user-defined input), re-name `filter_scale` to `filter_factor`, and then nondimensionalize all of the `IRREGULAR` grids by requiring their `grid_vars` to be nondimensional using the dimensional length of the smallest grid cell. I don't particularly like this. Seems like it requires extra and unnecessary work from the user.

Mainly I'd like some feedback on how to implement proposed change number 2, but I'd also welcome other ideas or general feedback about this issue.

NoraLoose:
[...]

PR #146, opened 2022-02-15
Fix dimensionality

iangrooms:
This PR addresses the problem of dimensionality described in #136. A quick summary: If a user tries any of the `REGULAR` Laplacians and uses a value of `dx_min` other than 1, they will get wrong answers. This PR will allow users to put in *dimensional* values of `dx_min` and `filter_scale` (in any system of units) and still get the right answers.

I am attempting to be minimally invasive. I've added a boolean class attribute `is_nondimensional` to all of the Laplacians. If the Laplacian is nondimensional, then at each step of the filter (in `filter.py`) it re-dimensionalizes by dividing by `dx_min ** 2`. The `filter_func`'s don't have access to `dx_min` so I added `dx_min_sq = dx_min ** 2` to the filter spec.

I initially started with @NoraLoose's suggestion (from #136) of defining new Laplacian classes, but as I worked on it it seemed like it would require bigger changes. For example, `_create_filter_func` accepts a `BaseScalarLaplacian` and I didn't want to have to define four new versions of `_create_filter_func` (scalar vs vector; dimensional vs non).

I also added a Note in the Basic Filtering section of the docs, saying that any `grid_vars` should have units that are consistent with `filter_scale` and `dx_min`, but that the specific system of units doesn't matter.

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -162,9 +162,20 @@ def _compute_filter_spec(
     s = np.array([y for x in s for y in x])
     is_laplacian = np.abs(s.imag / s.real) < root_tolerance
 
-    dx_min_sq = dx_min ** 2
+    dx_min_sq = dx_min**2
 
     return FilterSpec(n_steps_total, s, is_laplacian, s_max, p, n_iterations, dx_min_sq)
+
+
+def _maybe_dimensionalize(data, dx_min_sq):
+    """The point of this function is to avoid the expense of
+    dividing a large array by 1 when dealing with nondimensional
+    Laplacians
+    """
+    if dx_min_sq == 1:
+        return data
+    else:
+        return data / dx_min_sq
 
 
 def _create_filter_func(
@@ -194,8 +205,10 @@ def _create_filter_func(
                 if filter_spec.is_laplacian[i]:
                     s_l = np.real(filter_spec.s[i])
                     tendency = laplacian(field_bar)  # Compute Laplacian
-                    if Laplacian.is_nondimensional:
-                        tendency /= filter_spec.dx_min_sq  # dimensionalize
+                    if not Laplacian.is_dimensional:
+                        tendency = _maybe_dimensionalize(
+                            tendency, filter_spec.dx_min_sq
+                        )  # dimensionalize
                     field_bar += (1 / s_l) * tendency  # Update filtered field
                 else:
                     s_b = filter_spec.s[i]
@@ -203,9 +216,13 @@ def _create_filter_func(
                     temp_b = laplacian(
                         temp_l
                     )  # Compute Biharmonic (apply Laplacian twice)
-                    if Laplacian.is_nondimensional:
-                        temp_l /= filter_spec.dx_min_sq  # dimensionalize
-                        temp_b /= filter_spec.dx_min_sq ** 2  # dimensionalize
+                    if not Laplacian.is_dimensional:
+                        temp_l = _maybe_dimensionalize(
+                            temp_l, filter_spec.dx_min_sq
+                        )  # dimensionalize
+                        temp_b = _maybe_dimensionalize(
+                            temp_b, filter_spec.dx_min_sq**2
+                        )  # dimensionalize
                     field_bar += (
                         temp_l * 2 * np.real(s_b) / np.abs(s_b) ** 2
                         + temp_b * 1 / np.abs(s_b) ** 2
@@ -250,9 +267,13 @@ def _create_filter_func_vec(
                     (utendency, vtendency) = laplacian(
                         ufield_bar, vfield_bar
                     )  # Compute Laplacian
-                    if Laplacian.is_nondimensional:
-                        utendency /= filter_spec.dx_min_sq  # dimensionalize
-                        vtendency /= filter_spec.dx_min_sq  # dimensionalize
+                    if not Laplacian.is_dimensional:
+                        utendency = _maybe_dimensionalize(
+                            utendency, filter_spec.dx_min_sq
+                        )  # dimensionalize
+                        vtendency = _maybe_dimensionalize(
+                            vtendency, filter_spec.dx_min_sq
+                        )  # dimensionalize
                     ufield_bar += (1 / s_l) * utendency  # Update filtered ufield
                     vfield_bar += (1 / s_l) * vtendency  # Update filtered vfield
                 else:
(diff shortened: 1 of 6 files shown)


gcm_filters/filter.py
```diff
@@ -252,10 +263,15 @@ def filter_func_vec(ufield, vfield, *args):
...
-                    )
+                    if Laplacian.is_nondimensional:
+                        utemp_l /= filter_spec.dx_min_sq  # dimensionalize
+                        vtemp_l /= filter_spec.dx_min_sq  # dimensionalize
+                        utemp_b /= filter_spec.dx_min_sq ** 2  # dimensionalize
+                        vtemp_b /= filter_spec.dx_min_sq ** 2  # dimensionalize
+                        ufield_bar += (
+                            utemp_l * 2 * np.real(s_b) / np.abs(s_b) ** 2
+                            + utemp_b * 1 / np.abs(s_b) ** 2
+                        )
```
NoraLoose, line 274: Lines 271-274  should not be part of the if-statement.

rabernat (commented):
It's great to see this fix Ian! I can see that this resolves the original bug.

I'd like to brainstorm a bit whether there is a simpler way to do this, which doesn't require us to carry around this `is_nondimensional` attribute with all of the Laplacians.

Looking at the code I am struck by the fact that the `is_nondimensional` condition seems to somehow be closely tied to the value of `dx_min`. _Would it be possible to simply define a nondimensional Laplacian as a Laplaican for which `dx_min==1`?_ Are there scenarios that would break this assumption.

A corollary of this is the suggestion that perhaps we want to _attach dx_min to the Laplacian / grid, rather than to the Filter itself_. After all, dx_min is a property of the grid itself, not the filter.

Thanks for considering these comments. I'm just starting to really dig into this, but I know you have already through this through deeply.

NoraLoose:
Great work, Ian! 

If we stick to the `is_nondimensional` attribute, we should add some tests to this PR. In an ideal scenario, we want to test every newly added line to the code. Based on this, I can think of 2 tests:

1. A test that checks that the `is_nondimensional` attribute is there for each Laplacian, and has the expected value (e.g., to make sure a future PR does not accidentally change `is_nondimenional` from `True` to `False` for a Laplacian). Such a test could be added to `test_kernels.py` and could follow this structure:
```
@pytest.mark.parametrize(Laplacian, expected)
    ...

def test_nondimensional(Laplacian, expected):

    laplacian = Laplacian()
    assert laplacian.is_non_dimensional == expected
```
2. A test for the added lines in `filter.py`. For example, we could test if the filter gives the same thing if we put in values for `dx_min`, `area` etc. in cm vs. km.

Let me know if you would like help with these tests. But obviously, before implementing these test cases in `test_*`, we should agree on how we want to implement the code changes in `kernels.py` and `filter.py`.

gcm_filters/filter.py
iangrooms: Good catch! That would have been bad.

iangrooms:
Replying to @rabernat: 
> Would it be possible to simply define a nondimensional Laplacian as a Laplacian for which `dx_min==1`?

The nondimensional Laplacians don't just have _minimum_ grid spacing 1; they have every cell being a unit square. Identifying regular Laplacians with `dx_min==1` could break if, for example, there was a variable grid whose minimum was coincidentally 1 km (or 1 m or 1 cm, etc). The user would set `dx_min=1` and then the code would mistakenly think it was using a regular/nondimensional Laplacian.

> perhaps we want to attach `dx_min` to the Laplacian / grid, rather than to the Filter itself

The Filter does need to know about `dx_min`. The Filter object obtains a polynomial approximation that is accurate over a range of wavenumbers, with the highest wavenumber in the range set by `dx_min`. 

> perhaps we want to attach `dx_min` to the Laplacian / grid, rather than to the Filter itself

As noted above, the filter needs `dx_min`. But maybe your suggestion points to a different approach. My proposed fix has the same effect as changing the grid size in the regular Laplacians from 1 to `dx_min`, but without having to change the actual Laplacian kernel code. Instead of having the Laplacian kernel know about the grid size (which would be natural), I hacked `_create_filter_func` to fix the output of the nondimensional Laplacians. Not sure this is the best way, but it does seem to work. 

An more-natural alternative would be to add `dx_min` (or maybe just `dx`) to the `grid_vars` for the `REGULAR` Laplacians, and then change the code of the Laplacian kernels so that they divide by `dx ** 2` right before returning the result. I didn't choose this path because it would break all of the code that currently uses the `REGULAR` Laplacians, in addition to requiring a fairly major overhaul of `kernels.py`. If there's a strong argument for doing it this way even if it breaks backwards compatibility, then I'm happy to code it up. Curious to hear what you think.

gcm_filters/filter.py
NoraLoose: Wondering why the tests did not catch this? :confused:

iangrooms:
I added some tests as suggested by @NoraLoose. In `test_kernels.py` I add a test that just checks if the `is_nondimensional` attribute has the correct value. In `test_filters.py` I run the `REGULAR` filter on test data with `dx_min=1`, `filter_scale=4` and again with `dx_min=2` and `filter_scale=8`, and make sure the results are the same.

iangrooms:
The reason code coverage decreased is that I implemented the fix for non-dimensional vector Laplacians, but we don't have any non-dimensional vector Laplacians, so those lines don't get touched by the tests.

rabernat:
Hi Ian! Thanks for continuing to work on this PR. I have been a bit overwhelmed with getting ready for / participating in OSM for the past two weeks. Next week I intend to spend some time reviewing this PR in detail. (I am still confused about some parts of this and need to take some time to really understand what is happening.) Thanks for your patience.

rabernat (commented):
This is great Ian. Sorry for taking so long to review this. I really appreciate your patience.

A minor nit, I would prefer `dimensional = True`, rather `is_nondimensional = False`, since i find double negatives to be confusing. If you agree, you can change the name and invert the parameter. But this is annoying and not that important, so feel free to ignore that suggestion.

The only real suggestion I have is a minor performance optimization.

gcm_filters/filter.py
```diff
@@ -191,13 +194,18 @@ def filter_func(field, *args):
                 if filter_spec.is_laplacian[i]:
                     s_l = np.real(filter_spec.s[i])
                     tendency = laplacian(field_bar)  # Compute Laplacian
+                    if Laplacian.is_nondimensional:
+                        tendency /= filter_spec.dx_min_sq  # dimensionalize
```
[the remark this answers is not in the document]
iangrooms: Oh, I see. I thought it was the if statement that was the problem.

I think your fix is cleaner than mine, so let's go with that. I can work on both the double negative and the `maybe_nondimensionalize` and push something later today

iangrooms:
Had some problems related to this https://github.com/psf/black/issues/2964. My changes to the pre-commit configuration can be rolled back once this is fixed.

PR #153, opened 2022-05-10
Filter verification tests

NoraLoose:
This PR adds verification tests for the filter algorithm, in preparation for the refactor in #149, which replaces the filter algorithm with a new one. 

This PR essentially mirrors #79, but here the verification tests are associated with the `test_filter.py` module (rather than the `test_kernels.py` module as in #79). The main changes are:

- I moved fixture definitions from `tests/test_filter.py` to `tests/conftest.py` and, in the process, got rid off a lot of redundant code.
- I added the actual verification tests via `test_filter_validation.py`.


tests/test_filter.py
```diff
@@ -120,184 +120,14 @@ def test_filter_spec(filter_args, expected_filter_spec):
...
-            "kappa_iso": da_kappa,
-            "kappa_aniso": da_kappa,
-        }
-    data_u = np.random.rand(ny, nx)
-    data_v = np.random.rand(ny, nx)
-    da_u = xr.DataArray(data_u, dims=["y", "x"])
-    da_v = xr.DataArray(data_v, dims=["y", "x"])
-
-    return grid_type, da_u, da_v, grid_vars, geolat_u
-
```
rabernat, line 298: Can you just help me understand why all this code is being removed? Is it redundant with the code in conftest.py?

rabernat (commented):
Overall this look like a great improvement! Big :+1: to PRs that remove code! 😄 

I just have one question to make sure I understand what is happening here.

tests/test_filter.py
NoraLoose: Yes, exactly. The only two differences between the removed code and the code in `conftest.py` were:

1. `conftest.py` created numpy array test data (specific to `test_kernels.py`), whereas the removed code created xarray DataArray test data (specific to `test_filter.py`). The new fixtures `grid_type_and_input_ds` and `vector_grid_type_and_input_ds` in `conftest.py` take care of the numpy array--> xarray DataArray conversion, but without all the redundancy.
2.  The random number generator for the test data. Using the explicit constant seed for the random number generator in `conftest.py` ensures that we are always testing on the same data - a pre-requisite for the filter verification tests, exactly as done in #79.

NoraLoose:
@rabernat do you agree that we can merge this PR?

PR #157, opened 2022-08-18
Pin version of numcodecs

NoraLoose:
This is an attempt to close #156.

PR #158, opened 2022-09-08
Fix python version for tests

jbusecke:
As pointed out correctly by @pbranson [here](https://github.com/zarr-developers/numcodecs/issues/350#issuecomment-1228887635) the CI was actually not testing with the python versions specified (and seems to have defaulted to some newer version). 
This should correct that behavior.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,6 +33,7 @@ jobs:
           mamba-version: "*"
           channels: conda-forge, defaults
           channel-priority: true
+          python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


jbusecke:
With the pinned numcodec version from #157 this was failing for py3.6 (and passing for all others). I have removed the pin to see if that helps to resolve this. 
Apologies for not noticing this bug earlier BTW. I hope it does not cause too much disruption. I guess dropping support for python 3.6 could actually be reasonable.

NoraLoose:
Nice - thanks for looking into this @jbusecke! I will merge this.

PR #161, opened 2023-07-20
Relax constraint on kappa grid variable

NoraLoose:
The `IRREGULAR_WITH_LAND` takes `kappa_w` and `kappa_s` grid variables that allow to vary the filter scale spatially. There was a hard constraint on these kappas that they have to be equal to 1 somewhere in the domain. I softened the constraint to only checking that these kappas are very close to 1 somewhere in the domain (with a tolerance of 1e-05).


NoraLoose:
This PR may solve issue #162, too.

iangrooms (approved):
This looks good to me.

iangrooms:
Thanks @NoraLoose for noticing and fixing this, and thanks @marimpacheco for opening #162.

PR #174, opened 2024-11-12
Make sure `n_steps_default` >= 3

NoraLoose:
## Issue

Our current computation of the `n_steps_default` returns a value smaller than 3 for very small filter factors (<= 1.8) together with a Gaussian filter shape. If the user does not specify their own `n_steps`, the code will then set 
```
n_steps = n_steps_default < 3
```
which then leads to an error when the mass matrix for the Galerkin basis is set up (which assumes `n_steps` >=3).

## Changes

This PR simply makes sure that `n_steps_default` is always at least 3 by setting `n_steps_default` as the maximum of our current computation and 3. There is also a new test, which would have failed with the previous implementation.

### DIFF
--- a/gcm_filters/filter.py
+++ b/gcm_filters/filter.py
@@ -71,6 +71,24 @@ _target_function = {
 }
 
 
+def _compute_n_steps_default(
+    ndim, filter_shape, filter_scale, dx_min, transition_width
+):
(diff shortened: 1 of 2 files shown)


iangrooms (approved):
Looks good to me. Thanks for finding and fixing this.

